Issue #6, opened 2014-08-14
When message size goes out of sync, client hangs and does nothing


sbberic:
The nats server seems to be sending incorrect message size in the message "header". When the client receives a payload in stream.on('data') callback, it uses regular expressions to try and match for the header type (MSG, OK, ERR, PONG, etc) When the message size provided by the server is incorrect, it takes part of the next message and puts it as part of the current message. During the AWAITING_MSG_PAYLOAD step it incorrectly slices payload size so that on the next message inevitably falls through all the regular expression matches in AWAITING_CONTROL step and does nothing. 

e.g
Message 1: "MSG my.topic 2 10 {foo:bar}"
Message 2: "MSG my.topic 2 9 {bar:foo}"

Message 1 is interpreted as  "MSG my.topic 2 10 {foo:bar}M"
client goes out of sync and does nothing.

This seems like a server issue as well as a client one. The client probably should throw some error when this happens and the server needs to send the correct message size.

We're running gnatsd 5.0 and it's definitely not the latest so I was wondering if there was an issue similar to this that is already fixed. 

Our messages are JSON and varies between 15k to 80k in size

PR #7, opened 2014-10-14
Fixed buffer concatentation

abstractvector:
Fixing the Buffer concatenation to avoid character encoding issues, as per #6

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -388,6 +388,7 @@ Client.prototype.closeStream = function() {
     this.pSize     = 0;
     this.connected = false;
   }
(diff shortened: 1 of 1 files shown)


abstractvector:
It's not a case of a performance impact - the functionality before wasn't working, and now it is.

derekcollison:
Did you check why the tests failed?

derekcollison:
Looks like the tests got broken, so we need to fix that..

derekcollison:
There is a utf8 test which works correctly as is (at least under 0.10 & 12).. Does not need Buffer.concat.

PR #10, opened 2015-03-07
Basic cluster support

palmerabollo:
This is a first approach for basic cluster support in node-nats.

```url``` option can be a string or an array of strings (I've kept the same name but it could be easily renamed to ```servers```). For example:
```
var options = {};
options.url = [
 'nats://user1:pass1@server1',
 'nats://user1:pass1@server2:4243'
];
var nats = require('nats').connect(options);
```
If the client is not able to connect to a server, or the server goes down (stream 'close' event is received), it tries to reconnect to a randomly chosen server. This is a simple approach, that distributes load among servers, and can lead to reconnection attempts to the server that failed.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1,6 +1,6 @@
 /*!
  * Nats
- * Copyright(c) 2011-2014 Derek Collison (derek.collison@gmail.com)
+ * Copyright(c) 2011-2015 Derek Collison (derek.collison@gmail.com)
  * MIT Licensed
  */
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: lib/nats.js [hunk 1]
```diff
@@ -173,51 +175,66 @@ Client.prototype.parseOptions = function(opts) {
       options.url = DEFAULT_PRE + opts.port;
     }
     // Pull out various options here
-    this.assignOption(opts, 'url');
-    this.assignOption(opts, 'uri', 'url');
-    this.assignOption(opts, 'user');
```
palmerabollo: I've removed user and pass/password because they could lead to incompatible information, such as:
```
url: ['nats://userA:passA@localhost:4222'], user:'userB', pass:'passB'
```
But we could think about other policies (i.e. user and pass in the uri override the options)

derekcollison:
Have this functionality now, thanks..

PR #16, opened 2015-08-09
Miscellaneous updates for tests, linting and scripts.

cmfatih:
**Important Changes**

- Using `gnatsd` instead `nats-server` for local tests
- Update `.travis.yml` for `gnatsd`
- Update `.gitignore`, remove `.npmignore` and add `.istanbul.yml`
- Update for linting
- Let `examples` and `tests` to be in package.
- Comment ourt unused variables


derekcollison:
I have been wanting to do this, needed gnatsd to support TLS. Do I not have an auth test for Node? Thanks BTW..

cmfatih:
I didn't understand the auth question. But I know the auth tests are fine ( https://travis-ci.org/nats-io/node-nats/jobs/74835765#L200 ) Are you talking about adding TLS support (with tests) to `node-nats`? Which means `gnatsd` has already TLS support. OR are you talking about waiting until `gnatsd` supports TLS?

derekcollison:
Talking about TLS support, I was not sure if I tested that from the node
client. The ruby server had it (albeit not too secure), but gnatsd does not
yet.

cmfatih:
Oh I see... yeah there is no test for TLS atm.

BTW I'm using this library in [seneca-nats-transport](https://github.com/cmfatih/seneca-nats-transport) and it works great so far. Thanks for both the server and the client library. I will keep continue to bothering you with PRs :)

derekcollison:
That's great! Thanks.. Left a few comments, for most part looks good.

cmfatih:
@derekcollison : There are 2 errors in tests. Sometimes they pass, sometimes they fail.

```
# It happens in iojs

1) Queues should allow queue subscribers and normal subscribers to work together:
      Uncaught AssertionError: expected 2 to be 4
      + expected - actual
      -2
      +4
      
      at test/queues.js:60:23
      at Socket.<anonymous> (lib/nats.js:9:14302)
      at readableAddChunk (_stream_readable.js:146:16)
      at Socket.Readable.push (_stream_readable.js:110:10)
      at TCP.onread (net.js:521:20)
```

```
1) Cluster should randomly connect to servers by default:
     AssertionError: expected 39 to be within 40..60
      at Context.<anonymous> (test/cluster.js:68:23)
```

let me check those issues...

derekcollison:
See if the first is a timing thing.. The second is just a random number
thing, that one is ok IMO. We could increase the bounds..

cmfatih:
Okay I increased the range for 2nd issue. I'm gonna check the 1st one...

cmfatih:
Okay... I ran tests with iojs on my local env and can't reproduce it. When I have a time I will revisit tests and improve them. Currently all the tests are passing. https://travis-ci.org/nats-io/node-nats/builds/74937830

derekcollison:
Does the iojs one pass intermittently as well, or fail majority of the time?

cmfatih:
The iojs error is happened a few times on Travis. I couldn't reproduce locally (OSX). I'm not sure whether it is an iojs or gnatsd issue.

```javascript
  it('should allow queue subscribers and normal subscribers to work together', function(done) {
    var nc = NATS.connect(PORT);
    var received = 0;
    nc.subscribe('foo', {'queue':'myqueue'}, function() { received += 1; });
    nc.subscribe('foo', function() { received += 1; });
    nc.publish('foo');
    nc.publish('foo');
    nc.flush(function() {
      received.should.equal(4);
      nc.close();
      done();
    });
  });
```

derekcollison:
ok should not be a timing issue if I read the code right unless events can
fire out of order.. Besides this are we good to go? I will pull in, bump
version and push to npm. Thanks again..

cmfatih:
I have suspicious that server is getting killed during tests since Mocha runs those parallel. That's why I changed all the ports and make them unique for per test file. I will spend more time on it later. For now you can publish it if it is okay for you.

derekcollison:
ok I will pull it in.. Thanks..

PR #20, opened 2015-09-03
Adding assignOption to port fixes #19

### LINKED ISSUE #19: Cannot Specify External Connection
When I try to do:
```
nats.connect({
    url: '127.0.0.1',
    port: '4222'
  });
```

I get:
```
TypeError: port should be a number or string: null
    at Socket.connect (net.js:914:13)
    at Object.exports.connect.exports.createConnection (net.js:92:35)
```

It looks like there is no default for opts.port and opts.port is not being assigned as an option.

dlabey:


derekcollison:
If the url properly has a port and you specify one separately, does that override? Should we have a quick test to confirm that behavior?

dlabey:
I can specify one in the url so this is not as bad as I thought. Would it hurt though to allow the port to be specified in the options though?

derekcollison:
Again if they are specified in both places who wins?

dlabey:
Ah yes I see now.

PR #21, opened 2015-09-10
noDelay and WebSocket

ronag:
Assuming the API for the future WebSocket support in NATS looks the same I see no reason why not add web socket support into the clients. Until the server supports web sockets users can use a proxy server.

This code requires that one uses either browserify or webpack to get the required "Buffer" polyfill. Alternatively one can add a dependency to the "buffer" npm package.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -174,6 +174,7 @@ Client.prototype.parseOptions = function(opts) {
     'verbose'              : false,
     'pedantic'             : false,
     'reconnect'            : true,
+    'noDelay'              : false,
     'maxReconnectAttempts' : DEFAULT_MAX_RECONNECT_ATTEMPTS,
     'reconnectTimeWait'    : DEFAULT_RECONNECT_TIME_WAIT,
     'encoding'             : 'utf8'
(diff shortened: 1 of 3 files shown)


derekcollison:
Websocket support will most likely have a separate client, although similar API.

ronag:
@derekcollison: Could you explain why? Seems to me like just making necessarily complicated. As show n by this PR the solution is quite trivial...

derekcollison:
Fair comment. Two reasons..

1. The NATS protocol is optimized for simplicity and performance for low fanout (<100s) with a medium to large memory footprint per connection. WS would have different design constraints, very large fanout with low memory footprint and cpu per connection, etc. May require separate proto and separate server.

2. Identity and Security. Anything like this that will touch the *outside* world will most likely need a different server component with a different view of identity and auth.

shadiakiki1986:
The closest NATS WS client is [nats-websocket](https://github.com/simulacre/nats-websocket), but the last commit was 5 years ago. Can anyone point me in the direction of a NATS client for a browser? Or am I going down the wrong lane here? Should I just stick with a [pusherjs](https://github.com/pusher/pusher-js) client? There's an [amqp-js](https://github.com/dansimpson/amqp-js) client, but it uses flash :confused:

aricart:
@shadiakiki1986 Take a look at https://github.com/joshglendenning/websocket-nats, if that doesn't fit your use case. If that doesn't fit, let us know. Thanks!

PR #23, opened 2015-09-14
Allow user to configure client to attempt unlimited reconnects

ajschlosser:
We had an issue today where restarting our gnatsd server caused all of our clients eventually to give up trying to reconnect. Eventually, we had to restart many different processes across several different servers. It may be useful or desirable to allow users to configure their clients to make unlimited reconnect attempts (rather than some arbitrary large number, like 1000 or 10000).

If that is something that might be desired, adding an additional conditional to the ``close`` event listener would solve the issue. This way, a user would be able to pass in a ``maxReconnectAttempts`` value of ``-1`` in the client options to indicate that the client should attempt to reconnect indefinitely. (Some other value would of course work just as well...)

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -321,7 +321,7 @@ Client.prototype.createConnection = function() {
     client.emit('disconnect');
     if (client.closed === true ||
         client.options.reconnect === false ||
-        client.reconnects >= client.options.maxReconnectAttempts) {
+        ((client.reconnects >= client.options.maxReconnectAttempts) && client.options.maxReconnectAttempts !== -1)) {
       client.emit('close');
     } else {
       client.scheduleReconnect();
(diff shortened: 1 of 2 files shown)


derekcollison:
Why did they give up? How much time did it take to restart the new gnatsd? Was the MaxReconnectAttempts set or was it the default?

Thanks for the feedback.

Can we have some tests with the PR that test the functionality?

ajschlosser:
Everything was default--no indication of any kind of malfunctioning. IIRC, that means that it probably tried 10 times, once every 2 seconds, and gave up after 20 seconds. Just might be handy to be able to have it try forever (as opposed to some arbitrary number of times) for those of us who like to live on the wild side.

I'll be glad to contribute some tests to the PR. Thanks for your quick response!

derekcollison:
Update the PR with some tests, looks fine to me.

ajschlosser:
Test added and all tests passing.

PR #50, opened 2015-12-15
Unwrap `stream.resume()`

### LINKED ISSUE #49: Use of `stream.isPaused()` extraneous and breaks Node 0.10.x compatibility
The execution of `resume()` [implies](https://github.com/nodejs/node/blob/master/lib/_stream_readable.js#L700) the check to see if the stream is paused.  Thus, wrapping the `resume()` [here](https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L598) is unnecessary.

Moreover, using `isPaused()` breaks compatibility with Node v0.10.x which, while unarguably old, is required for my use case, presently.

Ulexus:
As `isPaused()` is already implied by `resume()`, its call is
extraneous.  This wouldn't be a problem, but this breaks compatibility
with NodeJS 0.10.x, which, while old, is still in the ecosystem.

Fixes #49

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 sudo: false
 
 node_js:
+  - "0.10"
   - "0.12"
   - "5.1"
   - "5.2"
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -595,9 +595,7 @@ Client.prototype.processInbound = function() {
   var start;
 
   // unpause if needed.
-  if (client.stream.isPaused()) {
-    client.stream.resume();
-  }
+  client.stream.resume();
 
   /* jshint -W083 */
 


Ulexus:
It looks like there are some non-trivial TLS problems.  Not an issue for my use case, but it doesn't seem like it is worth trying to maintain a broken version officially.  I can just fork a local copy, and _hopefully_ Meteor will update their NodeJS version soon.

meteor/meteor#5124

derekcollison:
Fair enough, i think with the changes you made it should work for non-TLS correct? Or should I merge this one to get it there?

Ulexus:
The latter commit (enabling the 0.10.x testing), you don't want.  Several of the TLS tests are broken in 0.10.x.

The first commit is trivial:  it just removes the redundant `isPaused()` call.

Yes, it works fine without TLS on 0.10.x.  I'm testing it now for my [Meteor wrapper](https://github.com/CyCoreSystems/meteor-nats).

PR #52, opened 2015-12-16
Add callback to sendCommand/flushPending

### LINKED ISSUE #51: Oddly affiliated callbacks for publish
The present scheme of processing callbacks for the publish function is a bit strange.  It is roughly akin to:  "I have tried to send your message.  I don't know whether it was sent or not, but after I threw it away, I was able to ping the server and get a pong back."

I feel that a more appropriate use of the callback is:  "I have tried to send your message to the server.  The send [failed|succeeded]."  

This says nothing about the operation of the server... but then, previous message was only a surrogate indicator, anyway.  If the server was unavailable for your message but came up in time to respond to the subsequent 'ping', your app is blithely unaware.  It could indirectly be listening for errors or connection losses, but even then, you won't know if that affected the message you sent or not.

Instead, I think the `publish` method's callback should be very specific in meaning, scoped to only the things it can do something about:  decide whether to try again.

This is strictly my opinion, and I'm very interested in hearing alternative views, particularly since I am rather new to NATS.

Ulexus:
This changes the semantics of the `publish` method's callback execution
to strictly mean that the message asked to be published was sent over
the wire (or failed to have been ).

Fixes #51 -- but also significantly changes the semantics (as witnessed
by the required changes to the tests)

As a side note:  there appears to have been an oversight in the test
'Basics should do callback after publish is flushed' where the actual
publish was not being sent, but the callback was being executed by
the following ping.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -518,13 +518,13 @@ Client.prototype.closeStream = function() {
  * @api private
  */
 
-Client.prototype.flushPending = function() {
+Client.prototype.flushPending = function(cb) {
   if (this.connected === false ||
       this.pending === null ||
       this.pending.length === 0 ||
       this.infoReceived !== true) { return; }
 
-  this.stream.write(this.pending.join(EMPTY));
+  this.stream.write(this.pending.join(EMPTY),cb);
   this.pending = [];
   this.pSize   = 0;
 };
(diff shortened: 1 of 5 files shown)


Ulexus:
Here is where I'm talking about:
https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L829

The callback is simply added to the list of pong callbacks and a ping is immediately generated (via a call to `flush()`.

Ulexus:
This is really only the first pass.  I think there are further refinements which could be made with the higher order methods, as well... but this is a start.

Ulexus:
... not sure why that travis run failed.  I've checked that precise version.

Ulexus:
#53 obviates this PR

PR #53, opened 2015-12-18
Better error handing with context

derekcollison:
Deals with issue #51.

/cc @Ulexus

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -66,12 +66,10 @@ var VERSION = '0.5.0',
     EMPTY = '',
 
     // Errors
-    PUB_BAD_SUBJECT_ERR = new Error('Publish: Subject must be supplied'),
-    PUB_BAD_MSG_ERR = new Error('Publish: Message can\'t be a function'),
-    PUB_BAD_REPLY_ERR = new Error('Publish: Reply can\'t be a function'),
-    PUB_CONN_CLOSED_ERR = new Error('Publish: Connection closed'),
-    FLUSH_CONN_CLOSED_ERR = new Error('Flush: Connection closed'),
-    SUB_CONN_CLOSED_ERR = new Error('Subscribe: Connection closed'),
+    BAD_SUBJECT_ERR = new Error('Subject must be supplied'),
+    BAD_MSG_ERR = new Error('Message can\'t be a function'),
+    BAD_REPLY_ERR = new Error('Reply can\'t be a function'),
+    CONN_CLOSED_ERR = new Error('Connection closed'),
 
     // Pedantic Mode support
     //Q_SUB = /^([^\.\*>\s]+|>$|\*)(\.([^\.\*>\s]+|>$|\*))*$/, // TODO: remove / never used
@@ -792,13 +790,13 @@ Client.prototype.processMsg = function() {
 Client.prototype.flush = function(opt_callback) {
   if (this.closed) {
     if (typeof opt_callback === 'function') {
-      opt_callback(FLUSH_CONN_CLOSED_ERR);
+      opt_callback(CONN_CLOSED_ERR);
       return;
     } else {
-      this.emit('error', FLUSH_CONN_CLOSED_ERR);
+      throw(CONN_CLOSED_ERR);
     }
   }
-  if( this.pongs ) {
+  if (this.pongs) {
     this.pongs.push(opt_callback);
     this.sendCommand(PING_REQUEST);
     this.flushPending();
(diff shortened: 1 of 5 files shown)


Ulexus:
Looks great @derekcollison ; Thanks!

lib/nats.js
```diff
@@ -801,9 +817,12 @@ Client.prototype.flush = function(opt_callback) {
 
 Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
   if (!msg) { msg = EMPTY; }
+  if (typeof subject !== 'string') {
+    this.emit('error', PUB_BAD_SUBJECT_ERR);
+  }
```
ronag: Um, what? This is weird... Shouldn't this go into the callback?

ronag:
This entire PR is strange to me? What's going on?

lib/nats.js
Ulexus: Indeed; this is not a connection-oriented error.  This is exactly where the callback should be used and a case where a connection error should definitely _not_ be used.

Ulexus: And the two emissions below, too.

lib/nats.js
```diff
@@ -824,10 +843,13 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
   } else {
     psub = 'PUB ' + subject + SPC + opt_reply + SPC;
   }
+
   this.sendCommand(psub + Buffer.byteLength(msg) + CR_LF + msg + CR_LF);
 
   if (opt_callback !== undefined) {
     this.flush(opt_callback);
+  } else if (this.closed) {
+    this.emit('error', PUB_CONN_CLOSED_ERR);
```
Ulexus: I would say this emission should be handled by a primary connection lost emission; connection losses themselves aren't request-oriented.  Failures of requests _due_ to connection loss are.  Thus, the callback, and only the callback is an appropriate vehicle for error transmission here.

lib/nats.js
```diff
@@ -843,6 +865,9 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
  */
 
 Client.prototype.subscribe = function(subject, opts, callback) {
+  if (this.closed) {
+    this.emit('error', SUB_CONN_CLOSED_ERR);
```
Ulexus: Context -> callback; !connection

lib/nats.js
derekcollison: ok, so these should only be localized and delivered to the callback. Fair. What if the callback is not being provided?

lib/nats.js
derekcollison: So this one should not emit either? So if no callback is provided, then the user would not have an indication of an error? This seems to be forcing users into always providing callbacks which is not efficient and generally an anti-pattern IMO.

lib/nats.js
derekcollison: Callback here is for message delivery and handling. There is not error presented to that callback, and adding one now where status callbacks expect it, at the beginning, would break existing code.

lib/nats.js
Ulexus: In Node, I take that as a presumption that the calling party doesn't care about the transmission, so nothing happens.

Actual errors on the connection (which include loss and failures) are handled by the stream/connection/client eventemitter.  Those are events unrelated to requests.  Most of that is already handled for you by the stream and connection.

lib/nats.js
Ulexus: In Node, that _is_ the pattern.  Your error handlers and your process control _are_ your callbacks.  Node is extremely callback-oriented, and it is designed to efficiently toss arbitrary functions onto the event stack.  That's how it works.

Ulexus: This is actually much of the problem with your main read loop.  It is too large a chunk of looping code to run efficiently in a Node environment (not that the loop itself is inefficient).  Node, being single-threaded, needs the separation in order to properly handle multiple code executions.  This is why you are having to put in that cool-off timer;  you shouldn't have to do that.

lib/nats.js
Ulexus: Understood, on backward compatibility.  That's why I didn't touch `subscribe` in my PR.  I don't have a good backwardly-compatible solution.  I think it will take a separate method.

lib/nats.js
Ulexus: Also, callbacks are not a substitution for connection-oriented events and their monitoring.  That scope, however, is not generally within the interest of the the piece of code executing this request.

A typical pattern:
```node
var c = new Client();
c.on('error',function(err) {
  console.log("Cry havok!",err);
}

setTimer(function({
  c.publish("hello",function(err) {
    console.log("Oops; I failed to send:",err);
  });
},1000);
```

Completely separate interests for connection events versus request events.  In a real program, you would have entire structures of executions for each of these.  Listening on `error` from the viewpoint of the timer function would be useless.

lib/nats.js
derekcollison: Should we throw an exception?

derekcollison: Regarding the read loop, I could process each message individually, but even then event loop starvation is possible and the performance for applications that behave correctly would suffer. I think the yieldTime is a good alternative.

Ulexus: No, emitting an 'error' here will automatically throw an exception if it is not otherwise handled.  I think that is sufficient.

I spent some time looking over the read loop, and no, I don't see any good alternatives without a significant restructuring... and I don't have enough idea about the real performance characteristics of NodeJS to say whether that would be worth doing.  I know it doesn't fit with the usual layout, and synchronous loops are always dangerous in NodeJS, but I don't (yet?) have a good opinion on a reorganisation, there.  yieldTime works

derekcollison: Always interested in good ideas and I think you and @ronag have better stylistic understandings of Node.js.

For the emit though, didn't we say that we should not direct to the connection handler for localized events? The connection closed on is special since if the connection is closed, all event listeners have been removed and the error will be thrown anyway, no way to avoid that as the code is written today.

Ulexus: The _client_ object should be persistent through connection losses, though.  It should have its own eventemitter (as it does) to handle events across connections.

Yes, were I to design afresh, with no mind to backward compatibility, I would keep errors in-scope.  I don't think we can safely add an error argument to the callback, though.  (By Node convention, the error parameter is the _first_ parameter to the callback.)

Thus, I don't see any way to adapt this to localize events.  I suppose you _could_ add a second callback, but that starts to get a bit ridiculous.

derekcollison: If the connection is dropped and we are trying to reconnect these will all work with no errors. If you closed the connection explicitly and then make a call to publish, subscribe, flush etc. this will get triggered.

Ulexus: True;  in that case, it should trigger an 'error' as you have.  The user can decide whether he should handle that 'error' or let it throw an exception.

PR #55, opened 2016-01-06
fix when nats urls fail at first server,  subscription state is lost

roytan883:
when  there are A, B, C gnatsd host, close someone, like B
then client UA, UB, UC, one of them like UC may first connect to B when noRandomize:false
UC will fail connecto B, then it reconnect to A, but UC's subs not send.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -336,9 +336,7 @@ Client.prototype.connectCB = function() {
   this.reconnecting = false;
   this.reconnects = 0;
   this.flushPending();
-  if (wasReconnecting) {
-    this.sendSubscriptions();
-  }
+  this.sendSubscriptions();
 
   this.wasConnected = true;
   this.currentServer.didConnect = true;
--- a/test/connect.js
+++ b/test/connect.js
@@ -58,5 +58,40 @@ describe('Basic Connectivity', function() {
       done();
     });
   });
+  
+  it('should still receive publish when some servers are invalid', function(done){
+    var natsServers = ['nats://localhost:22222', uri, 'nats://localhost:22223'];
+    var ua = NATS.connect({servers: natsServers});
+    var ub = NATS.connect({servers: natsServers});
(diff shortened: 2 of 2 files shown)


roytan883:
ok, i add a new test commit "add test when some servers are invalid"
It works with my last commit changes. 
I don't know why The Travis CI build failed, Travis CI not use the two commit together?

derekcollison:
Make sure local ```make test``` works.

roytan883:
sorry that i'm not familar with travis ci.
But, local test works, i already run this test before commit.
````
mocha connect.js 


  Basic Connectivity
    ✓ should perform basic connect with port
    ✓ should perform basic connect with uri
    ✓ should perform basic connect with options arg
    ✓ should emit a connect event
    ✓ should emit error if no server available
    ✓ should still receive publish when some servers are invalid (201ms)
    ✓ should still receive publish when some servers[noRandomize] are invalid (201ms)


  7 passing (672ms)

````

derekcollison:
Run ```make``` and ```make test```

roytan883:
OK, after fix missing ; of jshint,  'make' and 'make test' can work locally:
````
  ․ Basic Connectivity should perform basic connect with port: 0ms
  ․ Basic Connectivity should perform basic connect with uri: 0ms
  ․ Basic Connectivity should perform basic connect with options arg: 0ms
  ․ Basic Connectivity should emit a connect event: 0ms
  ․ Basic Connectivity should emit error if no server available: 0ms
  ․ Basic Connectivity should still receive publish when some servers are invalid: 200ms
  ․ Basic Connectivity should still receive publish when some servers[noRandomize] are invalid: 202ms
````
But there is one failing, I think which is not my test case:
````
  79 passing (8s)
  1 failing

  1) Split Messages "before all" hook:
     Error: Server exited with bad code, already running? (1 / null)
````

derekcollison:
Its all good. Thanks.

PR #57, opened 2016-01-08
Add binary support

petercgrant:
This adds support for binary messages:

```javascript
nats.publish('foo', new Buffer('\x00\xf0\x9f\x92\xa9\x00', 'binary'));
```

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -543,10 +543,10 @@ Client.prototype.flushPending = function() {
       }
       pos += len;
   });
-  this.stream.write(buf);
-
   this.pending = [];
   this.pSize   = 0;
+  this.stream.write(buf);
+
   return result;
 };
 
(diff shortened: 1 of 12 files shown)


derekcollison:
The tests under travis are failing. Also the changes look inefficient and will affect performance substantially. See if you can get travis tests to pass.

derekcollison:
I think binary support is a nice to have. Will take these tests and see if I can come up with a solution that does not impact performance as much. Thank you though for the PR.

petercgrant:
I saw that you already merged some of this PR. Thank you for that. I added
a couple more tests and fixed some races in the existing tests, which were
sometimes failing spuriously for me locally. I narrowed the failures down
to a couple of issues: event ordering and message loss during reconnect.
I've added two explicit tests in reconnect.js for these cases, and this
branch passes those tests. I left some comments so it would be clearer why
I moved things around the way I did, but I'm happy to explain it further if
any of it is unclear. Thanks!

derekcollison:
I used the tests but not any of the code changes to nats.js due to the bad performance it introduced. Could you make the additional changes to the tests against master and abandon this PR?

petercgrant:
Sure. I merged master into this PR already and took the string writing
portions from master, but I can submit the additional changes as a
different PR. The only piece of the buffer writing I changed post-merge was
the buffer and string mixed write, which I had already changed to create a
single large buffer to do a single write because I seemed to have more test
failures with multiple writes in a flush. However as I mentioned I tracked
this to apparent issues with event ordering, and issuing multiple writes
probably just gave node more opportunities to interleave events during the
flush. Anyway, the code is here now if you're curious what I'm up to.
Thanks again for incorporating these changes so quickly!

derekcollison:
Thanks that would be good. Right now what I do is mark if the pending queue has a buffer at all, and if so try to determine if they are all buffers, mixed or just strings. Seemed to yield the best performance for mixed and all buffer case.

PR #62, opened 2016-01-11
Event ordering

petercgrant:
These changes ensure that the client sends auth information and subscriptions before attempting to publish. The connect/reconnect event is now emitted before any publish/flush callbacks are called. This also fixes a unit test issue where clients weren't being closed after their tests were finished running, which sometimes failed unrelated tests when reconnect attempts were exhausted.


petercgrant:
I encountered a [spurious failure](https://travis-ci.org/nats-io/node-nats/jobs/100985085#L382) in my last PR. Notice the port in the uncaught error is 1423, the port used for the basic test suite, but the error is reported under the binary test suite.

derekcollison:
LGTM, had one minor comment on the sendSubscriptions not buffering and sending direct, probably not a big deal. 


Do we know how this effects performance if at all?

derekcollison:
Thank you BTW, good stuff here.

petercgrant:
I ran the benchmarks on a node:5.4-onbuild Docker container and saw very little change. That is what I expected because this is mostly a change in how (re)connection setup works. Do you think it's worthwhile to test reconnection time when the client has boatloads of subscriptions? I can put something together.

derekcollison:
I think it should not hold up this PR, but would be a good idea IMO.

derekcollison:
Thanks again for the PR!

PR #67, opened 2016-02-02
Added option to encode a JSON object before publishing a message and …

diogogmt:
To wrap the grunt work involved in encoding and decoding JSON messages I've added a new attribute named json to the client options.
The consequences of turning the JSON attribute to true would be:
Eencoding is set to base64
-JSON object is converted to a string before being published
-Base64 strings are encoded to JSON after being received as a message in its respective subscribers

I've also added a test to the Basic suite validating the new attribute.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -220,6 +220,7 @@ Client.prototype.parseOptions = function(opts) {
     this.assignOption(opts, 'name');
     this.assignOption(opts, 'client', 'name');
     this.assignOption(opts, 'yieldTime');
+    this.assignOption(opts, 'json');
   }
 
   var client = this;
@@ -234,7 +235,6 @@ Client.prototype.parseOptions = function(opts) {
   } else {
     throw new Error('Invalid Encoding:' + options.encoding);
   }
-
   // For cluster support
   client.servers = [];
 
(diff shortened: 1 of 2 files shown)


derekcollison:
What affect does this have on performance?

diogogmt:
@derekcollison as far as performance I don't see any hits since the application will have to do the encoding/decoding regardless.
The idea is to let the nats-client handle the encoding/decoding if JSON is the data format used by the application and all messages will be using the same format.
By default the JSON option can be set to false and the users can turn it on only when it sees fit.

derekcollison:
We should have it set to false by default, but even then interested if the code affects performance.

diogogmt:
Some notes on the review comments:

* If the message provided by the user is not a JSON object a BAD_JSON_MSG_ERROR exception is raised.
* As far as the encoding, it defaults to utf8 and when the subscriber receives a message it uses the same encoding set by the user.
* Removed the ascii toString encoding type when generating a string from the message buffer before parsing to a JSON object.


For the performance comparison, can we land a another PR with the benchmarks?

derekcollison:
Benchmarks are included, so you can test them now with your change and before.

Issue #68, opened 2016-03-13
TypeScript Definitions


xogeny:
The latest versions of TypeScript can pull typescript definitions from an NPM repository.  This helps
ensure that the TypeScript definitions are always in sync with what is published in node.  Furthermore,
it makes it much easier for TypeScript users to use the node module because all the have to do is
'npm install nats' and they automatically get TypeScript definitions.  With Angular2's adoption of TypeScript (and TypeScript's rising popularity, in general), this level of support for TypeScript will make life easier for a lot of developers.

All this means, for the `nats` developers, is including **one file** in their package.  It doesn't force any of you users to know or use TypeScript.  It just provides a definition file that can be found automatically by the TypeScript compiler.

Would you accept a pull request for such a definition file to be redistributed with the `nats` package in `npm`?

derekcollison:
Feel free to send a PR, happy to take a look.

tutman96:
I have created pull request #77 to address this issue. Please let me know if I did this process correctly.

tutman96:
Pull request #77 fixes this issue. It is available in v0.6.4. This can be closed now.

derekcollison:
Thanks.

PR #77, opened 2016-05-16
Fix for #68 adding typescript definition files

tutman96:


derekcollison:
We have typescript in the master tree I believe. Does this add something different? Happy to take a look.

tutman96:
I couldn't find any typescript in this project. The index.d.ts file and the small addition to the package.json shouldn't affect anything else in the repository however it will allow Typescript users to have typings for this module.

PR #79, opened 2016-05-23
add addServer(uri)

1N50MN14:
test added to `test/connect.js`

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -876,6 +876,17 @@ Client.prototype.processMsg = function() {
 };
 
 /**
+ * Push a new cluster server.
+ *
+ * @param {String} uri
+ * @api public
+*/
+
+Client.prototype.addServer = function(uri) {
+  this.servers.push(new Server(url.parse(uri)));
+}
+
+/**
  * Flush outbound queue to server and call optional callback when server has processed
  * all data.
  *
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -888,7 +888,7 @@ Client.prototype.addServer = function(uri) {
   if (this.options.noRandomize !==  true) {
     shuffle(this.servers);
   }
-}
+};
 
 /**
  * Flush outbound queue to server and call optional callback when server has processed
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -884,6 +884,10 @@ Client.prototype.processMsg = function() {
 
 Client.prototype.addServer = function(uri) {
   this.servers.push(new Server(url.parse(uri)));
+
+  if (this.options.noRandomize !==  true) {
+    shuffle(this.servers);
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/nats.js
```diff
@@ -876,6 +876,17 @@ Client.prototype.processMsg = function() {
...
 
 /**
+ * Push a new cluster server.
+ *
+ * @param {String} uri
+ * @api public
+*/
+
+Client.prototype.addServer = function(uri) {
+  this.servers.push(new Server(url.parse(uri)));
```
derekcollison: What if it already exists in the list? Should we de-dupe? Should we check for any errors from parse?

test/connect.js
```diff
@@ -118,4 +118,15 @@ describe('Basic Connectivity', function() {
     }, 100 * 2);
   });
 
+
+  it('should add a new cluster server', function(done){
+    var natsServers = [uri]
+    var nc = NATS.connect({servers: natsServers});
+    nc.on('connect', function(client) {
+      client.addServer('nats://localhost:22223');
+      client.servers.length.should.equal(2);
```
derekcollison: We should add another check to make sure the contents of the array are right. Should we check order?

Also, we normally randomly order the list of servers unless you ask us not to, so we need to honor that I believe, so simple push probably not sufficient.

lib/nats.js
1N50MN14: I assumed you're leaving this to the developer since there are currently no checks in the constructor; but I could add checks. If server already exists, ignore. Invalid uri, throw error.

test/connect.js
1N50MN14: Ok will check the opts for this one and re-apply order if required

lib/nats.js
derekcollison: If no checks in the constructor, no worries. Could not remember off top of my head.

test/connect.js
derekcollison: Thanks. And we should add tests on content of the server array.

1N50MN14: Ok I see that random ordering happens at the constructor level https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L241 i.e. not upon connect / reconnect so I can run `shuffle` again based on `opts.noRandomize` unless `noRandomize`is passed as a second argument in which case it overrides the default behaviour. Makes sense?

derekcollison: I think that sounds reasonable. Only hesitation I have is that if we end up with an order that hs me picking the same server on a disconnect for the subsequent re-connect. But I think that is something we can live with. So just call to shuffle() if opts.NoRandomize is false.

1N50MN14:
Apologies for the delay (two sick kids at home). Ok test now checks for the content of `servers` and  `addServer` shuffles if applicable; looks ok now?

derekcollison:
Failing tests, run make lint locally, I think you forgot a semi-colon.

1N50MN14:
Ok all checks are passing now it's the closing bracket on the function (my lint settings locally are a bit different, this one is easy to spot though). I guess I know how to do this now next time I attempt to submit a PR thanks for you time this was a good tiny exercise  👍

PR #82, opened 2016-07-19
Added token support

Swampy821:
Added support for using token for nats authentication. Also updated readme to include docs for auth.

## Authentication
```javascript

// Connect with username and password in the url
var nc = NATS.connect("nats://foo:bar@localhost:4222");

// Connect with username and password inside object
var nc = NATS.connect({'url':"nats://localhost:4222", 'user':'foo', 'pass':'bar'});

// Connect with token in url
var nc = NATS.connect("nats://mytoken@localhost:4222");

// Connect with token inside object
var nc = NATS.connect({'url':"nats://localhost:4222", 'token':'mytoken'});

```

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -72,7 +72,8 @@ var VERSION = '0.6.4',
     BAD_REPLY_ERR = new Error('Reply can\'t be a function'),
     CONN_CLOSED_ERR = new Error('Connection closed'),
     BAD_JSON_MSG_ERR = new Error('Message should be a JSON object'),
-
+    BAD_AUTHENTICATION_ERR = new Error('User and Token can not both be provided'),
+    
     // Pedantic Mode support
     //Q_SUB = /^([^\.\*>\s]+|>$|\*)(\.([^\.\*>\s]+|>$|\*))*$/, // TODO: remove / never used
     //Q_SUB_NO_WC = /^([^\.\*>\s]+)(\.([^\.\*>\s]+))*$/, // TODO: remove / never used
@@ -220,6 +221,11 @@ Client.prototype.parseOptions = function(opts) {
   
   // Set token as needed if in options.
   client.token = options.token;
+
+  // Authentication - make sure authentication is valid.
+  if (client.user && client.token) {
+    throw(BAD_AUTHENTICATION_ERR);
+  }
 
   // Encoding - make sure its valid.
   if (Buffer.isEncoding(options.encoding)) {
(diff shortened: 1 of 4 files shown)


lib/nats.js
```diff
@@ -425,6 +435,9 @@ Client.prototype.sendConnect = function() {
     cs.user = this.user;
     cs.pass = this.pass;
   }
+  if (this.token !== undefined) {
+    cs.auth_token = this.token;
+  }
```
aricart: Because of the varadic nature of option configuration it becomes possible to supply pass and token. In the case that both pass and token are provided likely we should fail the client connection and request that either password or token be provided.

test/auth.js
```diff
@@ -62,3 +62,60 @@ describe('Authorization', function() {
...
+  });
+
+  it('should connect with proper credentials as server url', function(done) {
+    var nc = NATS.connect({'servers':[authUrl]});
+    nc.on('connect', function(/*nc*/) {
+      setTimeout(done, 100);
+    });
+  });
+
+});
```
aricart: We should add a negative test where password and token are provided and see it rejected.

lib/nats.js
```diff
@@ -273,11 +277,17 @@ Client.prototype.selectServer = function() {
...
+    if (auth.length !== 1) {
+      if (client.options.user === undefined) {
+        client.user = auth[0];
+      }
+      if (client.options.pass === undefined) {
+        client.pass = auth[1];
+      }
+    } else {
+      if (client.options.token === undefined) {
+        client.token = auth[0];
```
aricart: Likely empty token would be invalid (server would reject, so probably nothing else to do here

aricart:
Besides the small comments, this LGTM. Thank you!

lib/nats.js
Swampy821: I completely agree. Will update shortly

lib/nats.js
Swampy821: I am not sure if you are advising a change or not here. If you are can you clarify more?

Swampy821:
I am confused why it is failing a test in travis-ci. I have ran the test multiple times using node 6.2.2 locally and not experiencing the error. It is also formatted almost identical to the other tests. 

Any advice?

lib/nats.js
aricart: I was just pointing out that `client.token` could be be null/empty.  In either case the server will reject as invalid credentials. (no action item).

aricart:
I restarted the test. Looks like it passed.

aricart:
Merged.

PR #85, opened 2016-08-11
Update publish method in json mode

CKing:
In JSON mode, publish tries to set an undefined variable `str`. Using node v6 will throw a `ReferenceError: str is not defined` when trying to execute this method.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -991,12 +991,12 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
         throw(BAD_JSON_MSG_ERR);
       }
       try {
-        str = JSON.stringify(msg);
+        msg = JSON.stringify(msg);
       } catch (e) {
         throw(BAD_JSON_MSG_ERR);
       }
     }
-    this.sendCommand(psub + Buffer.byteLength(str) + CR_LF + str + CR_LF);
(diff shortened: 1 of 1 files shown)


aricart:
Thanks for the patch.
Prior to applying your patch, I wrote a small test, and found it working. If you see on the code, the 'var str' is always in the scope of the call. I am not sure how you ran into that error. If you have a test case where you can repeat the issue, I can take a look.

```javascript
var NATS = require ('../'),
nsc = require('./support/nats_server_control'),
should = require('should');

describe('JSON payloads', function() {

  var PORT = 1423;
  var server;

  // Start up our own nats-server
  before(function (done) {
    server = nsc.start_server(PORT, done);
  });

  // Shutdown our server
  after(function () {
    server.kill();
  });
  
  it('should pub/sub with json', function(done){
    var nc = NATS.connect({json: true, port: PORT});
    var sid = nc.subscribe('foo', function(msg, reply, subj, sid){
      should.ok(typeof msg !== 'string');
      should.exist(msg.field);
      msg.field.should.be.equal('hello');
      should.exist(msg.body);
      msg.body.should.be.equal('world');
      nc.unsubscribe(sid);
      done();
    });

    nc.publish('foo', {field:'hello', body:'world'});
  });
});
```

CKing:
Never mind then.

It seems that line got lost during the checkout of a project. I force refetched the module and it works now again. I am sorry for the confusion (of myself).

PR #88, opened 2016-08-11
Move coveralls reporting to Travis after_success script

mcqueary:
Coveralls reporting was not working. This PR slightly changes the way we generate coverage reports, and the way we push the results to coveralls.

- Use `npm cover` to generate coverage locally rather than `make test-cov`.
- `npm coveralls` now executes `npm run cover -- --report lcovonly && cat ./reports/coverage/lcov.info | coveralls`
- `.travis.yml` now runs `npm run coveralls` in the `after_success`


aricart:
LGTM

Issue #89, opened 2016-09-04
 Change error vars (e.g. BAD_SUBJECT_ERR) to strings


shaneiseminger:
Working out an issue implementing the node-nats-streaming library and I was initially confused by the stack trace until I realized that it's because the `Error` objects are being instantiated as soon as the library was required, causing the stack trace to show the trace to the point the objects are instantiated, not the point that they're thrown.

Obvious suggestion is to set the error vars to strings and instantiate `Error` when throwing, e.g. `throw(new Error(BAD_SUBJECT_ERR))`. This change also seems slightly more efficient, as right now you have error objects always hanging around in memory whether they're needed or not.

Before this change, the stack trace looks like this:

````
<path>/node_modules/node-nats-streaming/node_modules/nats/lib/nats.js:938
      throw(BAD_SUBJECT_ERR);
      ^
Error: Subject must be supplied
    at Object.<anonymous> (<path>/node_modules/node-nats-streaming/node_modules/nats/lib/nats.js:70:23)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (<path>/node_modules/node-nats-streaming/node_modules/nats/index.js:1:80)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (<path>/node_modules/node-nats-streaming/lib/stan.js:15:12)
    at Module._compile (module.js:409:26)
````

After the suggested change:

````
Error: Subject must be supplied
    at Client.publish (<path>/node_modules/node-nats-streaming/node_modules/nats/lib/nats.js:938:13)
    at Client.request (<path>/node_modules/node-nats-streaming/node_modules/nats/lib/nats.js:1116:8)
    at Stan.subscribe (<path>/node_modules/node-nats-streaming/lib/stan.js:490:11)
[ stack trace from my code truncated ]
````

As you can see, the second stack trace is informative, the first is not.

Happy to submit a pull request for this unless you want to handle it a different way.

aricart:
[...]

PR #90, opened 2016-09-05
[FIX] https://github.com/nats-io/node-nats/issues/89 

aricart:
Stacktraces are not meaningful because Error objects are created with the library. Library now simply adds the error messages and constructs the error where thrown.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -23,7 +23,7 @@ var net    = require('net'),
  * Constants
  */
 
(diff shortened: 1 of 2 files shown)


aricart:
@ColinSullivan1 @kozlovic just looking for another pair of eyeballs on these simple changes.

ColinSullivan1:
LGTM

kozlovic:
LGTM

PR #91, opened 2016-09-07
unsubscribe #request inbox after callback call

jmventura:
Otherwise the #subs map will grow forever and overflow over time

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1119,6 +1119,7 @@ Client.prototype.timeout = function(sid, timeout, expected, callback) {
  */
 
 Client.prototype.request = function(subject, opt_msg, opt_options, callback) {
+  var self = this;
   if (typeof opt_msg === 'function') {
     callback = opt_msg;
     opt_msg = EMPTY;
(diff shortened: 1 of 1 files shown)


aricart:
@jmventura, the request function takes an `opt_options` where you can specify the number of responses before auto-unsubscribing. While typically one thinks of request/response as a 1-1 operation, there are many patterns where you want more than one.

In your use case where you want only one, you should be calling it like this:

`var sid = nc.request('foo', 'data', {max:1}, callback);`

Note that the subscription will not auto-unsubscribe until you get the number of responses you have specified, you can remedy that by using the `nc.timeout(sid, 1000, 1, callback);` and then unsubscribing if you don't get the number of messages you desired if the callback is triggered.

lib/nats.js
```diff
@@ -1131,6 +1132,8 @@ Client.prototype.request = function(subject, opt_msg, opt_options, callback) {
   var inbox = createInbox();
   var s = this.subscribe(inbox, opt_options, function(msg, reply) {
```
aricart: the `opts_options` can specify the number of responses to receive before auto-unsubscribing. See the jsdoc above.

jmventura:
Hi Alberto, cool, would you please illustrate the _"other patterns where you want more than one"_. I can't imagine them yet.

Thanks a lot,
Juan

aricart:
Control plane applications. Where you use messaging to manage a number of other processes. Maybe you want to get the least busy list of processes. If your repliers 'slow' responses based on how busy they are, the first N to respond will indicate the least busy. Google type queries also follow this approach. A single request goes to thousands of servers. But only some of the responses are used. NATS is actually optimized for that type of pattern. Where you do make a request to many (read thousands) but only want a small number of responses back.

jmventura:
Great explanation, thanks again Alberto.

PR #95, opened 2016-11-01
support raw buffers with {encoding: null}

barn3y:
# Overview

Allow the use of raw Buffer objects with node-nats, by passing `{encoding: null}` as an explicit option. Note that the default encoding, UTF8, is not changed.


barn3y:
Whoops - this is supposed to be on a fork, not the original project. Closing.

Issue #97, opened 2016-11-19
request doesn't support rpcTimeOut


sfrooster:
Unlike csharp and java, a "request/reply" action doesn't seem to allow for a timeout after which the request is auto-failed.

aricart:
Sure it does! The timeout is not specified with the request. Instead there's a timeout function that you can call and associate with the subscriber id (returned by subscribe and request). Here's an example.

```
#!/usr/bin/env node

'use strict';

var nats = require('nats').connect();

nats.on('error', function(e) {
  console.log('Error [' + nats.options.url + ']: ' + e);
  process.exit();
});

var sid = nats.request("foo", {'max':1}, function(e) {
  console.log('callback called: ' + e);
});

nats.timeout(sid, 1000, 1, function(sid){
  console.log('the subscriber timedout: ' + sid);
});

```

PR #99, opened 2016-11-21
addresses issue 97

sfrooster:
see #97

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -53,6 +53,10 @@ export interface SubscribeOptions {
 	max?: number
 }
 
+export interface RequestOptions extends SubscribeOptions {
+	timeout?: number
+}
+
 declare class Client extends events.EventEmitter {
 	/**
 	 * Create a properly formatted inbox subject.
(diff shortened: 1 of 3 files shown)


derekcollison:
Be good to have some tests to prove it works correctly. Also check why tests failed.

sfrooster:
I fixed he original round of failures, current failure only occurs with
Node6:

  1) Yield should yield to other events:

     Uncaught AssertionError: expected 21 to be within 10..20

      at Assertion.fail (node_modules/should/cjs/should.js:205:17)

      at Assertion.value [as within] (node_modules/should/cjs/should.js:277:19)

      at Timeout.<anonymous> (test/yield.js:33:20)


Don't know what that is.  As for a new test, my question is what behavior
would be inline with a timeout from the other clients?

derekcollison:
That failure may be an anomaly. The testing framework allows for async testing, so we should add in a test that passes only if the request times out within the right amount of time.

derekcollison:
Let me know when you add in the tests. If you could also rebase with master (not required) that would be great.

sfrooster:
Tried writing mocha tests to catch async errors thrown and treat as success - no luck.

sfrooster:
looks like there are other ways to tackle this in node

aricart:
The correct is way is https://github.com/nats-io/node-nats/issues/97

Issue #104, opened 2016-12-03
Errors should be instance of Error, not string


WoLfulus:
This gets really hard to handle exceptions without proper error codes, for example, I'd like to shutdown the application if there's a ETIMEDOUT error, and keep running on other codes. 

Check if the error string contains "ETIMEDOUT" isn't an option.

aricart:
What error are you getting exactly? This is likely an error coming from the socket. Where are you getting it. Can you share the stack trace?

WoLfulus:
I mean, the lib itself emits errors as strings, not Error instances.

https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L310

I can't rely on analysing string contents to check what errors are happening.

Example: 

```
this.emit('error', new SecureConnectionRequiredError());
```

aricart:
Let me see what  I can do. This change will create a backwards compatible issues for everyone else if they are relying on the string.

aricart:
@WoLfulus go ahead a see if the PR fits for you. In the case of the connection issue, instead of wrapping the error or adding additional properties, I am just bubbling the error up without any kind of augmentation. All errors generated by the library now have constants as well to aid in testing the errors. If there are any other issues, please feel free to reopen or create another issue.  PR will be merged once additional eyeballs land on it.

WoLfulus:
Where does the `code` of the error goes? Error class receives only one parameter on the constructor. The second one is lost.

aricart:
This is what I just did the source for Error gave me the impression that the error id was exposed. I implemented exactly what you propose. The NatsError also provides a `code`, `chainedError` so you can compare codes. If there's a connection error you can further inspect the `chainedError` for the reason. Thanks for the catch.

WoLfulus:
Perfect! Thanks :)

PR #105, opened 2016-12-03
[FIX] #104 - errors emitted by the library are now errors and not strings

aricart:
All errors thrown/emitted now have codes that can be used for testing the type of error raised. Errors emitted by the library are now errors and not strings. In the case of connection
errors, the original error raised by the OS is surfaced.

**If your client code depends on the value of the messages, you will have to update your code before upgrading to the library.** To aid in handing, the Error objects can be tested, but a more reliable way is provided by the new constants (BAD_SUBJECT, BAD_MSG, BAD_REPLY, CONN_CLOSED, BAD_JSON, BAD_AUTHENTICATION, INVALID_ENCODING, SECURE_CONN_REQ, NON_SECURE_CONN_REQ, CERT_REQ, NATS_PROTOCOL_ERR)

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -86,6 +86,8 @@ var VERSION = '0.7.0',
     NON_SECURE_CONN_REQ_MSG = 'Server does not support a secure connection.',
     CLIENT_CERT_REQ = 'CLIENT_CERT_REQ',
     CLIENT_CERT_REQ_MSG = 'Server requires a client certificate.',
+    CONN_ERR = 'CONN_ERR',
+    CONN_ERR_MSG_PREFIX = 'Could not connect to server: ',
     NATS_PROTOCOL_ERR = 'NATS_PROTOCOL_ERR',
 
     // Pedantic Mode support
@@ -94,10 +96,20 @@ var VERSION = '0.7.0',
 
     FLUSH_THRESHOLD = 65536;
 
+
+  var NatsError = function(message, code, chainedError) {
+    Error.captureStackTrace(this, this.constructor);
+    this.name = this.constructor.name;
+    this.message = message;
+    this.code = code;
+    this.chainedError = chainedError;
+  };
+  util.inherits(NatsError, Error);
+  exports.NatsError = NatsError;
+
 /**
  * Library Version
  */
-
 exports.version = VERSION;
 
 /**
@@ -254,14 +266,14 @@ Client.prototype.parseOptions = function(opts) {
 
   // Authentication - make sure authentication is valid.
   if (client.user && client.token) {
-    throw(new Error(BAD_AUTHENTICATION_MSG, BAD_AUTHENTICATION));
+    throw(new NatsError(BAD_AUTHENTICATION_MSG, BAD_AUTHENTICATION));
   }
 
   // Encoding - make sure its valid.
   if (Buffer.isEncoding(options.encoding)) {
     client.encoding = options.encoding;
   } else {
-    throw new Error(INVALID_ENCODING_MSG_PREFIX + options.encoding, INVALID_ENCODING);
+    throw new NatsError(INVALID_ENCODING_MSG_PREFIX + options.encoding, INVALID_ENCODING);
   }
   // For cluster support
   client.servers = [];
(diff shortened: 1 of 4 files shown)


derekcollison:
LGTM

lib/nats.js
```diff
@@ -66,12 +66,27 @@ var VERSION = '0.6.8',
     PONG_RESPONSE = 'PONG' + CR_LF,
 
     // Errors
-    BAD_SUBJECT = 'Subject must be supplied',
-    BAD_MSG = 'Message can\'t be a function',
-    BAD_REPLY = 'Reply can\'t be a function',
-    CONN_CLOSED = 'Connection closed',
+    BAD_SUBJECT = 'BAD_SUBJECT',
+    BAD_SUBJECT_MSG = 'Subject must be supplied',
```
wallyqs: Just an idea, but maybe there could custom error types?  Wondering whether something like this could be done for example:
```js
BAD_SUBJECT = 'Subject must be supplied';
BAD_SUBJECT_MSG = 'Subject must be supplied'
BAD_SUBJECT_ERR = function BAD_SUBJECT_ERR(err){
    Error.captureStackTrace(this, this.constructor);    
    this.name = this.constructor.name;
    this.message = BAD_SUBJECT_MSG;
};
util.inherits(BAD_SUBJECT_ERR, Error);
```

That way it would be possible to catch errors per type for example:
```js
try {
    // ...throws custom error
} catch (e) {
    if (e instanceof BAD_SUBJECT_ERR) {
        console.log("- Error: ", e.name);
        console.log("- Error: ", e.message);
        console.log("- Stack: ", e.stack);
    }
    if (e instanceof Error) {
        console.log("= Error: ", e.name);
        console.log("= Error: ", e.message);
        console.log("= Stack: ", e.stack);
    }
}
```

lib/nats.js
```diff
@@ -401,7 +444,7 @@ Client.prototype.setupHandlers = function() {
     // Only bubble up error if we never had connected
     // to the server and we only have one.
     if (client.wasConnected === false && client.servers.length === 0) {
-      client.emit('error', 'Could not connect to server: ' + exception);
+      client.emit('error', new NatsError(CONN_ERR_MSG_PREFIX + exception, CONN_ERR, exception));
```
wallyqs: Seems that the `chainedError` is only used for this case, maybe an alternative could be to just thrown the error opaquely instead of wrapping with new NatsError type? Think that way caller could catch the original type and nats errors without having to check further state on the nats error which was thrown.

aricart: The issue is the that event handler will be called, and now they have to test if the error has the property or not. With the current form, the errors are uniform, but in the case of the connection they can look to see if there's additional information.

lib/nats.js
```diff
@@ -737,7 +780,7 @@ Client.prototype.processInbound = function() {
       } else if ((m = OK.exec(buf)) !== null) {
         // Ignore for now..
       } else if ((m = ERR.exec(buf)) !== null) {
-        client.emit('error', m[1]);
+        client.emit('error', new Error(m[1], NATS_PROTOCOL_ERR));
```
aricart: good catch

WoLfulus:
LGTM also! Thanks for this!

aricart:
Pushing it when @wallyqs  or @derekcollison  do the final LGTM.

lib/nats.js
```diff
@@ -1007,7 +1050,7 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
   if (opt_callback !== undefined) {
     this.flush(opt_callback);
   } else if (this.closed) {
-    throw(new Error(CONN_CLOSED));
+    throw(new NatsError(CONN_CLOSED_MSG, CONN_CLOSED));
```
wallyqs: Maybe out of scope for this PR, but also another idea is that instead of using a code to determine the error, new error types based could be created which extend from `NatsError` could be added, like a `ConnectionClosedError` for example.

aricart: It becomes a switch by typeof or by code. More things to maintain considering that we have to then alias all those types for typescript.

lib/nats.js
wallyqs: OK I see just thought that maybe keeping the original type could be useful, but maybe better if the error types are uniform as you mention.

lib/nats.js
WoLfulus: This is what I suggested on #104 

Seems like a good approach too, instanceof can be used to handle the errors.

aricart: I would go for it, but with typescript it just doubles on the things we have to keep in sync.

wallyqs:
LGTM, maybe could help updating some of the tests or adding new ones which check against `NatsError` being thrown and also maybe adding new error types instead of basing on a code string to handle base on types instead, but that could be tried in another PR in separate.

aricart:
I added two tests checking that the error type does what it does. Later we can add more checks to the existing tests.

lib/nats.js
```diff
@@ -66,26 +66,69 @@ var VERSION = '0.6.8',
...
     // Pedantic Mode support
     //Q_SUB = /^([^\.\*>\s]+|>$|\*)(\.([^\.\*>\s]+|>$|\*))*$/, // TODO: remove / never used
     //Q_SUB_NO_WC = /^([^\.\*>\s]+)(\.([^\.\*>\s]+))*$/, // TODO: remove / never used
 
     FLUSH_THRESHOLD = 65536;
 
+
+  var NatsError = function(message, code, chainedError) {
+    Error.captureStackTrace(this, this.constructor);
+    this.name = this.constructor.name;
```
wallyqs: ah, I think for `this.constructor.name` to be set here you have to declare it like ` var NatsError = function NatsError(err){`, as it right now might just be empty

aricart: Absolutely correct. Good catch.

PR #110, opened 2017-01-10
Add new option "preserveBuffers"

StarpTech:
Add support for sending raw octet data
Add test

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -901,7 +901,7 @@ Client.prototype.processMsg = function() {
           if (this.options.preserveBuffers) {
             msg = JSON.parse(this.payload.msg.toString());
           } else {
-            msg = JSON.parse(new Buffer(this.payload.msg, this.options.encoding).toString());
+            msg = JSON.parse(this.payload.msg.toString(this.options.encoding));
           }
         } catch (e) {
           msg = e;
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -898,7 +898,11 @@ Client.prototype.processMsg = function() {
       var msg = this.payload.msg;
       if (this.options.json) {
         try {
-          msg = JSON.parse(new Buffer(this.payload.msg, this.options.encoding).toString());
+          if (this.options.preserveBuffers) {
+            msg = JSON.parse(this.payload.msg.toString());
+          } else {
+            msg = JSON.parse(new Buffer(this.payload.msg, this.options.encoding).toString());
+          }
         } catch (e) {
           msg = e;
         }
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -211,6 +211,7 @@ Client.prototype.parseOptions = function(opts) {
     this.assignOption(opts, 'yieldTime');
     this.assignOption(opts, 'waitOnFirstConnect');
     this.assignOption(opts, 'json');
+    this.assignOption(opts, 'preserveBuffers');
   }
 
   var client = this;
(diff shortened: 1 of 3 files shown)


aricart:
@StarpTech Would there be an issue here:
https://github.com/StarpTech/node-nats/blob/db4d2d5a4cc061704ad273de33236a265a6f52a9/lib/nats.js#L897

StarpTech:
@aricart 
Yes it works but can be improved. I added a test and improved the place.

StarpTech:
@aricart allright? Can you release it on npm?

derekcollison:
Can someone give me a breakdown of any performance changes?

StarpTech:
@derekcollison I can't have a performance lose. All is the same if you don't set `preverseBuffers: true` also with that flag the buffer will be passed directly to the consumer it must be even faster.

aricart:
@StarpTech I will not push it until I try a few things out. The code looks reasonable, but we will need to run some performance tests etc.

aricart:
@derekcollison 

Ran the JS performance tests, without changing how the payload is generated. The publisher for the sub_perf test is nats-bench. Each bench ran 6 times. Showing max/min values.

# Base numbers - without the PR
pub_perf.js 716332-658761 msgs/s
pub_sub_perf.js 504413-437349 msgs/s
reconnect_perf.js 8666-8080 ms
request_perf.js 13442-13276 req/res sec lat 37-36µs.
sub_perf.js 757575-702247 msgs/s


# PR but treating msgs with default encoding (connection not modified)
pub_perf.js 724375-625782 msgs/s
pub_sub_perf.js 501504-497388 msgs/sec
reconnect_perf.js 8497-8075 ms
request_perf.js 13462-13304 req/res sec lat 37-37µs.
sub_perf.js 761035-687757 msgs/s


# Tests with 'preserveBuffers' option set
pub_perf.js 721500-672947 msg/sec
pub_sub_perf.js 449943-432900 msgs/s
reconnect_perf.js 8643-8300 ms
request_perf.js 13292-13220 req/res sec lat 37-37µs.
sub_perf.js 604594-518672 msgs/s | ran again as this was expected better got 599520-579374 msgs/s

aricart:
@StarpTech Most of the benchmarks look reasonable. However the sub_perf.js which is actually the one expected to run faster doesn't. Could you identify why this is? - BTW, when running the subscriber test, I used nats-bench from go-nats to generate the messages like this: `nats-bench -np 1 -n 500000 -ms 0 test`

StarpTech:
Hi @aricart,  great summary but you have to also vary the payload size. I think that would make the test closer to real world scenrious. I also have to say that the Intention of this PR isnt a performance improvement it should only allow to transfer raw buffers end to end. This feature is configurable.

StarpTech:
I will try to identify the problem.

StarpTech:
@aricart which node version do you use?

StarpTech:
@aricart after some research it seems that heavily depends on the payload size. Could you start your benchmarks with a high payload size like 200-500KB? 

> Strings are built-in to V8, and allocate memory within the VM. Buffers were added not to make all string operations faster, but to represent binary data, where as Strings are unicode.

When writing large amounts of data to a socket, it's much more efficient to have that data in binary format, vs having to convert from unicode.

So for common operations, like concat, I'm not surprised Strings are faster.
Ref: http://stackoverflow.com/questions/4901570/buffer-vs-string-speed-why-is-string-faster

I will try some separate tests.

derekcollison:
The runtime will decide on its own if strings are faster, and once we preserve buffers this might run counter to what the v8 runtime would optimize for. These results looks reasonable to me.

LGTM

StarpTech:
# Subscriber Performance

Node: 6.9.2
CPU: i5-6600K 3.5GHz
SSD

I used pub_perf.js and sub_perf.js for testing.

**Payload size:** 1024 bytes
**Messages:** 500000

With preserveBuffers
Published at 215703-242013 msgs/sec

Without preserveBuffers, encoding: 'binary'
Published at 189465-206611 msgs/sec

**Payload size:** 5012 bytes
**Messages:** 500000

With preserveBuffers
Received at 346981-395256  msgs/sec

Without preserveBuffers, encoding: 'binary'
Published at 296559-323415 msgs/sec



**Payload size:** 5 bytes
**Messages:** 500000

With preserveBuffers
Published at 1075268-1103752 msgs/sec

Without preserveBuffers, encoding: 'binary'
Published at 1319261-1392757 msgs/sec

derekcollison:
Thanks..

derekcollison:
@aricart Bump version before push to npm..

aricart:
Just did that.- waiting on the build. Also updated readme

/a


On January 11, 2017 at 1:30:33 PM, Derek Collison (notifications@github.com)
wrote:

@aricart <https://github.com/aricart> Bump version before push to npm..

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/nats-io/node-nats/pull/110#issuecomment-271969809>, or mute
the thread

derekcollison:
Perfect thanks!

PR #114, opened 2017-01-24
fix invalid host reference

pkoretic:
`host` is found inside `url` object which is inside `currentServer` object


aricart:
LGTM. The parsed server is an Url.

aricart:
Thanks for the catch!

PR #115, opened 2017-01-25
trailing CR_LF will break binary protocols

StarpTech:


StarpTech:
I think the failing build for node 0.12 is related to this PR. Could be a estimation `delta.should.within(10, 25);` ?

StarpTech:
yep, it was :)

StarpTech:
@aricart @derekcollison would be awesome if you could review it.

aricart:
@StarpTech Can you write a negative test?
I augmented the binary.js, to do reuse the logic of the binary test, but also have a specific but taking into account that the payload can be a a buffer and it passes without your change. I did something like this for all tests, let me put a pr up so you can further write some tests

```
nc.subscribe('invalid2octet', function(msg) {
      msg.length.should.equal(2);
      if(nc.options.preserveBuffers) {
        should.ok(invalid2octet.equals(msg));
      } else {
        msg.should.equal(invalid2octet.toString('binary'));
      }
      finished();
    });
```

aricart:
Here's the branch with the passing test.

https://github.com/nats-io/node-nats/tree/binary-tests

StarpTech:
@aricart  The reason why it doesn't fail is because they are processed together not in chunks. Apart from this nowhere is CR_LF_LEN appended. You can test it if you do:

Change line 866 to 
```js
client.payload.msg = client.inbound.slice(0, client.payload.size + CR_LF_LEN);
```
Check `invalid2octet`
```js
    nc.subscribe('invalid2octet', function (msg) {
      console.log((new Buffer("\r\n")).toJSON()); //{ type: 'Buffer', data: [ 13, 10 ] } // binary data of \r\n
      console.log((msg).toJSON()); //{ type: 'Buffer', data: [ 195, 40, 13, 10 ] } see its appended
      msg.length.should.equal(2);
      if(nc.options.preserveBuffers) {
        should.ok(invalid2octet.equals(msg));
      } else {
        msg.should.equal(invalid2octet.toString('binary'));
      }
      finished();
    });
```

aricart:
BTW, if you do toJSON() at least on my side I get an error. Doing JSON.stringify() will correctly build. However the error doesn't show unless there's a very large message payload. I updated the test on the branch I have. If you can please incorporate the test into your PR, then I'll merge it.

(https://github.com/nats-io/node-nats/tree/binary-tests)

StarpTech:
Hi @aricart done. I also could reproduce it through test.

StarpTech:
@aricart can you retrigger? Thats not my test.

aricart:
@StarpTech Awesome - one more thing. Could you please squash all the changes and revert the change on the readme (noop). I'll merge after that goes through.

StarpTech:
@aricart sry but I can'´t rebase it. It freeze on my computer. Can you di it for me? Thanks.

StarpTech:
@aricart can you publish it today?

aricart:
working on it.

aricart:
@StarpTech it is available on npm.

Issue #121, opened 2017-02-14
Add Timeout and default max=1 to requests


MichaelErmer:
The GO implementation has a timeout option and sets max to 1 by default: https://github.com/nats-io/go-nats/blob/master/nats.go#L1848

We are using nats very heavily lately and had to find out about max:1 the hard way, after multiple server crashes at about ~3M subscriptions.

I think its unintuitive that there is no default max value, it would also be good to have a timeout option like GO to get rid of stale requests, ie when the processing server is crashing during the request.

aricart:
In node the current pattern is this:

```var nats = require('nats').connect();

ssid = nats.request('foo', '', {max:1}, function(msg) {
  console.log('response: ' + msg);
});
nats.timeout(ssid, 1000, 1, function(ssid) {
  console.log('ssid ' + ssid + ' timed out');
});```

The issue here is that while in your use case you want to unsubscribe as soon as you get one response, the code in the wild may be expecting N number of responses. So the change will break existing clients that upgrade.

PR #122, opened 2017-02-14
Add timeout to request, default max responses to 1 for requests

MichaelErmer:
see #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,14 +29,21 @@ nats.subscribe('foo', function(msg) {
 var sid = nats.subscribe('foo', function(msg) {});
 nats.unsubscribe(sid);
 
-// Request Streams
-var sid = nats.request('request', function(response) {
+// Request Streams, will Auto-Unsubscribe after 1 response
+var sid = nats.request('request', function(error, response) {
+  if (error) console.error(error);
   console.log('Got a response in msg stream: ' + response);
 });
 
-// Request with Auto-Unsubscribe. Will unsubscribe after
-// the first response is received via {'max':1}
-nats.request('help', null, {'max':1}, function(response) {
+// Request without Auto-Unsubscribe. Will never unsubscribe
+nats.request('help', null, {'max':false}, function(error, response) {
+  if (error) console.error(error);
+  console.log('Got a response for help: ' + response);
+});
+
+// Request with timeout, returns error if no response was received after 10000 ms
+nats.request('help', null, {'timeout': 10000}, function(error, response) {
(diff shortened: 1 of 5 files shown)


aricart:
As you can see from the tests, changing the API like this will create compatibility issues with existing clients.

README.md
```diff
@@ -29,14 +29,21 @@ nats.subscribe('foo', function(msg) {
...
+var sid = nats.request('request', function(error, response) {
+  if (error) console.error(error);
   console.log('Got a response in msg stream: ' + response);
 });
 
-// Request with Auto-Unsubscribe. Will unsubscribe after
-// the first response is received via {'max':1}
-nats.request('help', null, {'max':1}, function(response) {
+// Request without Auto-Unsubscribe. Will never unsubscribe
+nats.request('help', null, {'max':false}, function(error, response) {
```
aricart: this is a won't do. the callback function has a different signature. So existing clients will be called with an empty error argument.

lib/nats.js
```diff
@@ -1165,6 +1167,39 @@ Client.prototype.timeout = function(sid, timeout, expected, callback) {
...
+  if (typeof opt_msg === 'function') {
+    callback = opt_msg;
+    opt_msg = EMPTY;
+    opt_options = {};
+  }
+  if (typeof opt_options === 'function') {
+    callback = opt_options;
+    opt_options = {};
+  }
+  if (!opt_options.max && opt_options.max !== false) {
```
aricart: If we were to need a different flag, why overload the one that is there?

lib/nats.js
```diff
@@ -1184,17 +1219,37 @@ Client.prototype.request = function(subject, opt_msg, opt_options, callback) {
...
-    opt_options = null;
+    opt_options = {};
+  }
+  if (!opt_options.max && opt_options.max !== false) {
+    opt_options.max = 1;
   }
   var inbox = createInbox();
+  
+  // To keep old behaviour, remove after major version bump
+  function wrapCallback (error, msg, reply) {
```
aricart: Great that this handles the internal issue of calling the argument in a backward compatible way. But now when I go look in my IDE, I will see something that will confuse me. Because the first expected argument on the callback is an error. Yet by automagic it does something that doesn't match the signature of the api.

aricart:
I we are to do a single message with timeout api, we need the functionality to not alter the existing behaviors which are widely used, even if not exactly matching a different flavor of the api. I think there's merit in asking for such convenience, but we don't need to break existing clients.

PR #124, opened 2017-02-17
#123 - Merge metalog:fix/msg-is-array adding new test

aricart:
Fixes #123


aricart:
@metalog Thanks for the fix. I added a test and merged it. NPM is already updated.

Issue #126, opened 2017-02-17
unsubscribe can leave timeouts hanging


aricart:
Unsubscribe doesn't look if a subscription has an associated timeout. This means that is possible for a client that specified a timeout on a subscription to call unsubscribe, but the timeout will get to fire as it is never cancelled when the subscription is removed from the subscription list.

PR #127, opened 2017-02-17
[FIX] #126 - Unsubscribe can leave timeouts pending, and then getting invoked even after the subscription is cancelled.

aricart:


kozlovic:
LGTM

PR #131, opened 2017-03-02
[FIX] Add max argument. REQ_TIMEOUT null fix

eridem:
While I am testing this example, I got some errors:

- `NATS.REQ_TIMEOUT` was not defined.
- `requestOne` was expecting an object in the third place instead of the timeout.

This fixed the example for these specific cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ npm install nats
 ## Basic Usage
 
 ```javascript
-
-var nats = require('nats').connect();
+var NATS = require('nats');
+var nats = NATS.connect();
 
 // Simple Publisher
 nats.publish('foo', 'Hello World!');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ nats.request('help', null, {'max':1}, function(response) {
 
 
 // Request for single response with timeout.
-nats.requestOne('help', null, 1000, function(response) {
+nats.requestOne('help', null, {'max':1}, 1000, function(response) {
```
aricart: The requestOne doesn't require the max option, but accepts any option that can be passed to the underlying subscription. So the correct incantation for the sample should be: 
nats.requestOne('help', null, {}, 1000, function(response) {} or 
nats.requestOne('help, null, null, function(response){}
```

eridem: Understand 👍 . In any case the argument is missing (null or {}). I will do a quick modification.

eridem: The new commit should do the trick I believe 👍

PR #143, opened 2017-05-04
FIX #103 - Implementing discovered servers API

### LINKED ISSUE #103: Support Servers and DiscoveredServers APIs
See https://github.com/nats-io/go-nats/issues/214

aricart:
Extended nats_server_control to allow easy creation of a cluster
Bumped versions

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -23,7 +23,7 @@ var net    = require('net'),
  * Constants
  */
 
-var VERSION = '0.7.16',
+var VERSION = '0.7.18',
 
     DEFAULT_PORT = 4222,
     DEFAULT_PRE  = 'nats://localhost:',
@@ -482,7 +482,8 @@ Client.prototype.sendConnect = function() {
     'lang'    : 'node',
     'version' : VERSION,
     'verbose' : this.options.verbose,
-    'pedantic': this.options.pedantic
+    'pedantic': this.options.pedantic,
+    'protocol': 1,
   };
   if (this.user !== undefined) {
     cs.user = this.user;
@@ -494,7 +495,6 @@ Client.prototype.sendConnect = function() {
   if (this.options.name !== undefined) {
     cs.name = this.options.name;
   }
-
   // If we enqueued requests before we received INFO from the server, or we
   // reconnected, there be other data pending, write this immediately instead
   // of adding it to the queue.
@@ -797,6 +797,33 @@ Client.prototype.processInbound = function() {
 	if (client.checkTLSMismatch() === true) {
 	  return;
 	}
+
+  // Always try to read the connect_urls from info
+  if(client.info.connect_urls && client.info.connect_urls.length > 0) {
+    // don't add duplicates
+    var known = [];
+    client.servers.forEach(function(server) {
+      known.push(server.url.href);
+    });
+    // add new ones
+    var toAdd = [];
+    client.info.connect_urls.forEach(function(server) {
+      var u = 'nats://' + server;
+      if(known.indexOf(u) === -1) {
+        toAdd.push(new Server(url.parse(u)));
+      }
+    });
+
+    if(toAdd.length > 0) {
+      if(client.options.noRandomize !== true) {
+        shuffle(toAdd);
+      }
+      toAdd.forEach(function(s) {
+        client.servers.push(s);
+      });
+    }
+  }
+
 	// Process first INFO
 	if (client.infoReceived === false) {
 	  // Switch over to TLS as needed.
--- a/test/dyncluster.js
+++ b/test/dyncluster.js
@@ -1,0 +1,143 @@
+/* jslint node: true */
+/* global describe: false, before: false, after: false, it: false, afterEach: false, beforeEach: false */
+/* jshint -W030 */
+'use strict';
+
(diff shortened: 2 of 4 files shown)


lib/nats.js
```diff
@@ -797,6 +801,23 @@ Client.prototype.processInbound = function() {
 	if (client.checkTLSMismatch() === true) {
 	  return;
 	}
+
+  // Always try to read the connect_urls from info
+  if(client.info.connect_urls && client.info.connect_urls.length > 0) {
```
kozlovic: First, indentation looks strange.
Second, are you not going to do that for any protocol message received by the client? Looks to me that you should ensure that protocol being processed is an INFO here.
Also, not sure if you support randomization, but what the other clients do is randomize (if allowed) the array or connect_urls prior to add the URLs to the server pool.

lib/nats.js
```diff
@@ -797,6 +801,23 @@ Client.prototype.processInbound = function() {
...
+    // don't add duplicates
+    var known = [];
+    client.servers.forEach(function(server) {
+      known.push(server.url.href);
+    });
+    // add new ones
+    client.info.connect_urls.forEach(function(server) {
+      var u = 'nats://' + server;
+      if(known.indexOf(u) === -1) {
+        client.addServer(u);
```
derekcollison: Does this do the right thing with randomizing or not the list?

test/dyncluster.js
```diff
@@ -0,0 +1,81 @@
...
+              should(nc.servers.length).be.equal(3);
+              done();
+            }, 1000);
+          });
+        });
+      });
+    });
+  });
+
+  it('setting protocol zero doesnt update', function(done) {
```
derekcollison: Why do we want to expose and let client decide? Is that how we do it in other clients, they can suppress if they want to?

lib/nats.js
kozlovic: Just to be clear, we don't want to randomize the pool when we add things after the initial connect. We can randomize the received URLs *before* adding to the pool, but not randomize the entire pool after initial connect.

aricart: Yes it does, it is calling addServer which handles randomization options:
https://sourcegraph.com/github.com/nats-io/node-nats@b5db73ee06428471568a730c3b0ff7f907fa6943/-/blob/lib/nats.js#L990:18

test/dyncluster.js
```diff
@@ -0,0 +1,81 @@
...
+    servers = nsc.start_cluster([port], route_port, function() {
+      should(servers.length).be.equal(1);
+
+      // connect nats - protocol is set to zero to prevent the info updates
+      var nc = NATS.connect({'port': port, 'reconnectTimeWait': 100, 'protocol': 0});
+      nc.on('connect', function () {
+        // add two servers
+        process.nextTick(function () {
+          var others = nsc.add_member_with_delay([port + 1, port + 2], route_port, 250, function () {
+            should(others.length).be.equal(2);
```
derekcollison: We should write test to check randomize etc. And not randomize.

lib/nats.js
aricart: I preserved the indentation that was in the file :)
The block we are in is only handling INFO messages.  See https://sourcegraph.com/github.com/nats-io/node-nats@b5db73ee06428471568a730c3b0ff7f907fa6943/-/blob/lib/nats.js#L798:1

test/dyncluster.js
aricart: I can remove the option.

lib/nats.js
kozlovic: Even with this link, I am sorry but it looks horrible. It is very difficult to know what is in which block.

aricart:
@kozlovic @derekcollison I have put in the changes. `protocol` option is removed. Shuffling of the discovered servers is done and then appended to the current list and an associated test.

derekcollison:
LGTM, let's have @kozlovic give a +1 as well.

lib/nats.js
```diff
@@ -797,6 +797,31 @@ Client.prototype.processInbound = function() {
...
+    var toAdd = [];
+    client.info.connect_urls.forEach(function(server) {
+      var u = 'nats://' + server;
+      if(known.indexOf(u) === -1) {
+        toAdd.push(new Server(url.parse(u)));
+      }
+    });
+
+    if(toAdd.length > 0) {
+      shuffle(toAdd);
```
kozlovic: If I am not mistaken, this client has a `noRandomize` option. If so, you should shuffle the incoming array only if `noRandomize === false`.

aricart:
@kozlovic I made the shuffling of the discovered servers honor noRandomize, and added a related test.

kozlovic:
LGTM. Not sure why we get so many coveralls report (I deleted some from this page).

aricart:
There were some git errors during checkout for one of the builds.

/a


On May 5, 2017 at 1:15:27 PM, Ivan Kozlovic (notifications@github.com)
wrote:

LGTM. Not sure why we get so many coveralls report (I deleted some from
this page).

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nats-io/node-nats/pull/143#issuecomment-299565082>, or mute
the thread

Issue #145, opened 2017-05-11
Error when reconnecting to server & Attempting to send messages


blazedd:
We've encountered a conditional bug where the NATS client will attempt to send messages, despite no longer being connected to the server. In the code it's not checking to see if the socket is still attached or not and casuses an error. The side affect appears to prevent the application from completing a reconnect as well. Once completely restarted it can once again reconnect to the server.

You can see our code usage here: https://github.com/nodecraft/appframe-nats.js

```
[14:34:40] [LOG]: [NATS] Lost connection from server. Reconnecting... at appframe._log (~/node_modules/appframe/appframe.js:882:15)
[14:34:40] [ERROR]: Cannot read property 'resume' of null
TypeError: Cannot read property 'resume' of null
    at Client.processInbound (~/node_modules/nats/lib/nats.js:761:16)
    at Socket.<anonymous> (~/node_modules/nats/lib/nats.js:468:12)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:547:20) at appframe.debug (~/node_modules/appframe/appframe.js:864:15)
```

aricart:
@blazedd can you attach the version number as written in the nats.js.

blazedd:
@aricart It uses the latest `0.7.18` that's available on NPM.

aricart:
@blazedd Also are you getting called on 'close', 'disconnect' or 'error'.

blazedd:
@aricart We are only seeing `reconnecting` and `error` events.

aricart:
@blazedd Can you give me your connect options.
the only way that the stream is nullified is by closeStream(). 

If we are raising an error, the error argument ill have additional information. It has several properties. Can you print them out. ie:
```
on('error', function(ne) {
 console.log('error name', ne.name);
 console.log('error message', ne.message);
 console.log('error code', ne.code);
 console.log('error chained error', ne.chainedError);
});
```

blazedd:
Connect options:
```
"url": "nats://REDACTED:4222",
"user": "REDACTED",
"password": "REDACTED",
"json": true,
"tls": true,
"maxReconnectAttempts": -1
```

Error:
```
name: 'NatsError',
message: '\'Authorization Timeout\'',
code: 'NATS_PROTOCOL_ERR',
chainedError: undefined
```

[...]

PR #146, opened 2017-05-14
Fix Reconnect Memory Leak

blazedd:
This resolves issue #145

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -561,8 +561,6 @@ Client.prototype.createConnection = function() {
   this.stream = stream;
   // Setup the proper handlers.
   this.setupHandlers();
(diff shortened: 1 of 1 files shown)


aricart:
Ouch! - I have almost the same fixes that you have there. With additions on handling other error conditions that I am about to push as a PR.

aricart:
I have merged https://github.com/nats-io/node-nats/pull/147 which implements a very similar fix to the one proposed here. In addition to these fixes it, the PR addresses behavioral issues so that the node client matches other nats client when a protocol errors happens.

Note that on the original error addressed, in this version the client will still `close`, but you will get a `close` notification. The fix for the `authorization timeout` is to increate the timeout on the server configuration.

PR #147, opened 2017-05-15
[FIX] #145 - when using TLS reconnects would fail because of listener leaks (events firing on leaked streams)

aricart:
[FIX] node client fired error on any protocol error. Some protocol errors such as 'stale connection' should have initiated a reconnect.
[FIX] if a client attempted to publish/subscribe to a subject for which it had no permission, an error was generated. New event ('permission_error') will now be sent to the client.

@blazedd effectively submitted a [similar PR](https://github.com/nats-io/node-nats/pull/146).

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -23,7 +23,7 @@ var net    = require('net'),
  * Constants
  */
 
-var VERSION = '0.7.18',
+var VERSION = '0.7.20',
 
     DEFAULT_PORT = 4222,
     DEFAULT_PRE  = 'nats://localhost:',
@@ -91,6 +91,8 @@ var VERSION = '0.7.18',
     NATS_PROTOCOL_ERR = 'NATS_PROTOCOL_ERR',
     REQ_TIMEOUT = 'REQ_TIMEOUT',
     REQ_TIMEOUT_MSG_PREFIX = 'The request timed out for subscription id: ',
+    STALE_CONNECTION_ERR = "stale connection",
+    PERMISSIONS_ERR = "permissions violation",
 
     // Pedantic Mode support
     //Q_SUB = /^([^\.\*>\s]+|>$|\*)(\.([^\.\*>\s]+|>$|\*))*$/, // TODO: remove / never used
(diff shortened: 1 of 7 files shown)


aricart:
Note that I have included 'authorization timeout' as a non-fatal error. Other clients don't handle this, we can remove this behaviour or enhance other clients to reconnect if a server fails to authorize because of some internal load issue. Not sure if this is the right thing to do.

derekcollison:
The server should be expected to disconnect after an authorization timeout. I would cleanup the connection and fire the event callbacks.

aricart:
Fire 'error'?

derekcollison:
Yes I believe so..

aricart:
Updated so the client disconnects on the timeout error. Added another test without tls which will always get the error from the server.

lib/nats.js
```diff
@@ -91,6 +91,9 @@ var VERSION = '0.7.18',
     NATS_PROTOCOL_ERR = 'NATS_PROTOCOL_ERR',
     REQ_TIMEOUT = 'REQ_TIMEOUT',
     REQ_TIMEOUT_MSG_PREFIX = 'The request timed out for subscription id: ',
+    STALE_CONNECTION_ERR = "stale connection",
+    PERMISSIONS_ERR = "permissions violation",
+    AUTHORIZATION_TIMEOUT_ERR = "authorization timeout",
```
kozlovic: Not used.

test/perms.js
```diff
@@ -0,0 +1,148 @@
...
+        ncu.writeFile(cf, ncu.JsonToYaml(conf));
+        server = nsc.start_server(PORT, ['-c', cf], done);
+    });
+
+    // Shutdown our server
+    after(function () {
+        nsc.stop_server(server);
+    });
+
+    it('foo can subscribe/pub foo', function (done) {
```
kozlovic: I would not test the positive cases here, this is a server test, not a client. You want to test a part of the code that handles the failure, so I would keep only negative tests where we try to pub/sub and get an error and check that connection is not closed.

lib/nats.js
aricart: good catch

kozlovic:
LGTM

Issue #144, opened 2017-05-05
Add jslinter to Travis


petemiron:
Add a linter (JSHint or ESLint) to travis builds to ensure consistent formatting and prevent common issues.

cc @kozlovic

aricart:
@petemiron @kozlovic We actually do run a linter however formatting is not part of it.
I was thinking that likely we should use `jscs` - perhaps moving from jshint to eslint.
One thing is that changing the formatting will basically we'll dump all the blames on the file.

Also it we should pick some sort of style. I would go for https://github.com/jscs-dev/node-jscs/blob/master/presets/crockford.json, but realize that this will have significant changes on the code formatting, including variable declarations etc. Suggestions welcomed.

kozlovic:
That's something I love about Go is that the formatting is no subjective ;-)
I don't know if we need a tool or not, or just simply manually fixing the obvious ones such as the one where adding code for INFO protocol. In other words, when adding or updating some code, take the opportunity to fix the formatting.

StarpTech:
I vote for https://standards.com

PR #149, opened 2017-05-25
[FIX] #144 - Updated linter to eslint.

aricart:
[FIX] Reformatted sources with js-beautify
[FIX] Linter warnings
[CHANGE] Linter is only executed for node versions 4 and greater. As eslint makes use of features incompatible with node 0.12.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+node_modules/**
+reports/**
+.eslintrc.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,290 @@
+module.exports = {
+    "env": {
+        "node": true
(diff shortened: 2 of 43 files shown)


aricart:
@derekcollison - this is the reformatting PR and upgrade to eslint. The only salient note is that linters are run on travis only for node 4 or greater. If we should pick a different code style please let me know.

derekcollison:
Your call IMO. Solicit the team's opinion. I am ok with what you all propose.

ColinSullivan1:
@aricart , LGTM, readable, clean.  Elsewhere we discussed how this would affect blame; perhaps you can amend the commit to mention the git blame `-w` flag for user convenience.

PR #156, opened 2017-08-08
typescript changes for tls options

### LINKED ISSUE #155: typings for tls don't allow tls options (object)
Hi for my test server I am using ClientOptions in the form of:
`{
    user: 'myuser',
    pass: 'mypass',
    tls: {
          rejectUnauthorized: false
    }
}`

 and I am using TypeScript. Without manually changing Typescript gives me an error because tls only accepts a boolean. Can you add `| Object` at least?

aricart:
[ci skip]
[FIX #155] Modified typings definition to allow boolean | Object for tls option.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,6 @@
 
 import events = require('events');
+import tls = require('tls');
 
 export const version: string;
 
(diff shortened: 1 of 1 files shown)


tutman96:
I would suggest changing it from `Object` to `tls.TlsOptions`. It would take a `import tls = require('tls');` but then there is type completion for the various tls options.

aricart:
@tutman96 thanks for the suggesting. Can you take a look at the changes?

tutman96:
lgtm

aricart:
@tutman96  I have also opened an RFC in https://github.com/nats-io/node-nats/issues/157. Likely you have some insight and would like to comment. Thanks for the review.

PR #164, opened 2017-10-16
[FIX] Flapping Tests.

aricart:


test/support/nats_server_control.js
```diff
@@ -7,15 +7,34 @@ var net = require('net');
 var SERVER = (process.env.TRAVIS) ? 'gnatsd/gnatsd' : 'gnatsd';
 var DEFAULT_PORT = 4222;
 
+// select some random start port between 40000 and 50000
+var next = Math.floor(Math.random()*(50000-40000+1)+40000);
```
kozlovic: This does not guarantee that the server will be able to start if the port is in use. It looks like lots of tests are failing with connection refused, so not sure if this is because of that or something else.

aricart: open to suggestions on how to allocate a random port that works.

kozlovic: None. If you were to set to `-1` then server would pick a random non allocated port, but then from your client, you don't know the port (without doing some tricky things such as parse log, etc..). In go, it is easy since the client tests start the server from a go-routine they can get that generated port from the resulting server options.
The issue I have is that you are trying to solve flapping tests but it looks like this method is creating even more issues.
If there were flapping tests due to port issues, it is more likely that a test may not kill the server properly on failure (or success), or that some tests are executed in parallel (not sure if that's possible).

aricart: Will have to continue to study the problem. Likely will need to add additional clustering support to know if the cluster is formed correctly rather than rely on timing. I am going to defer that for now, as this makes it pass more often than it did before

kozlovic:
As discussed, you will have another pass to try solve specific clustering tests.

PR #166, opened 2017-10-20
[FIX] Made all server stops sync

aricart:
[FIX] Server stops made sync otherwise, test setup/destroy had a chance of interfering with the wrong process.

[FIX] Removed test that is effectively non-sensical callback test.

### DIFF
--- a/test/support/nats_server_control.js
+++ b/test/support/nats_server_control.js
@@ -15,11 +15,12 @@ function start_server(port, opt_flags, done) {
         done = opt_flags;
         opt_flags = null;
     }
-    var flags = ['-p', port];
+    var flags = ['-p', port, '-a', '127.0.0.1'];
 
     if (opt_flags) {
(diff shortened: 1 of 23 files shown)


test/support/nats_server_control.js
```diff
@@ -102,9 +103,24 @@ function start_server(port, opt_flags, done) {
...
+        }
+    } else {
+        setTimeout(function () {
+            wait_stop(server, done);
+        });
+    }
+}
+
+function stop_server(server, done) {
+    if (server) {
```
kozlovic: If `stop_server` is always called with a server, you should not test for `server` and let it fail hard so we can fix the test.

kozlovic: We looked at the test suite and there are cases with clustering that we end up calling stop with a null reference.

kozlovic:
LGTM

PR #162, opened 2017-10-06
Add REQUEST_TIMEOUT constant to typescript exports.

pettero:

PR #167, opened 2017-10-20
TS Exports

aricart:
Exporting some constants so they are available to typescript. Initial change was by @pettero (https://github.com/nats-io/node-nats/pull/162)


aricart:
Changes in the code are simply sorting the constants, and exposing them in the index.d.ts file.

PR #169, opened 2017-10-20
[BUMP] MAX_CONTROL_LINE_SIZE to 1024 to match server default

aricart:


lib/nats.js
```diff
@@ -27,7 +27,7 @@ var VERSION = '0.7.24',
     DEFAULT_PRE = 'nats://localhost:',
     DEFAULT_URI = DEFAULT_PRE + DEFAULT_PORT,
 
-    MAX_CONTROL_LINE_SIZE = 512,
+    MAX_CONTROL_LINE_SIZE = 1024,
```
petemiron: Do we need this at all? Can't we just use `client.inbound.length` on https://github.com/nats-io/node-nats/blob/cc523148b3eff03081e0f6343afd747822ee83e4/lib/nats.js#L767

wallyqs: Should probably check too how doing the slice for more bytes affects the performance as would always be doing a slice of 1024 bytes instead. Also maybe try out a way without using the max control line limit, haven't checked perf but could also do something like this in the parser (could be way slower but does not depend on max control line):
```js
	    var pos = client.inbound.indexOf(CR_LF);
	    if (pos == -1) {
		// No full control line yet so skipping
		return;
	    }
            var buf = client.inbound.toString('binary', 0, pos+CR_LF_LEN);
```

aricart: The issue here is that as Wally is pointed out, this is simply setting the size of the buffer. The bad part is the inbound buffer will continue to grow, but the processing will get stuck on the MAX_CONTROL_LINE_SIZE bytes without correctly parsing the message header, and repeat without doing anything until it runs out of memory.

IMHO, the first fix is to normalize the size to match server default as this PR is trying to do.

Then the discussion that likely needs to happen is whether the server advertises what MAX_CONTROL_LINE_SIZE and clients dynamically adjust. I would imagine that even if we added the ability to edit this default at the client level, that would somewhat pointless as the server can always get out sync with the client.

I also think that since the length of subjects, replyTo, and client ids greatly affect this value, our specification should state limits on these and then derive/enforce this value, as their length also have an impact on server performance.

Finally the message header is parsed as a binary string which may result in surrogate pairs perhaps creating incorrect strings.

ColinSullivan1: LGTM, as a stopgap measure.

aricart: @wallyqs The performance varies. On these set of runs they are fairly stable. Likely we would like to keep the default, and simply expose the setting to change it. The real fix is likely to do a zero allocation parser like go-nats.

```[ar@barsoom:~/go/src/git...-nats/benchmark]$ node pub_sub_perf_test.js -s 512 -i 10
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 532197 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 509035 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 508130 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 513874 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 512295 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 505944 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 505433 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 507872 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 510855 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=512 503524 msgs/sec
max: 532197  min: 503524 avg: 510915.9
[ar@barsoom:~/go/src/git...-nats/benchmark]$ node pub_sub_perf_test.js -s 1024 -i 10
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 504540 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 538213 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 529100 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 526177 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 504795 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 493949 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 485436 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 476190 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 467508 msgs/sec
Publish/Subscribe Performance Test (2000000) MaxControlLineSize=1024 468493 msgs/sec
max: 538213  min: 467508 avg: 499440.1```

ColinSullivan1:
Is this still on the table?

sgtram:
Guys, could someone say when this will be released?
We're experiencing a really critical issue when we stop receiving any messages.
When I tried to patch lib/nats.js and set MAX_CONTROL_LINE_SIZE=1024 it starts working again.
We're using the latest version of the server 1.1.0 and 1.0.4

aricart:
If your server size is not matching the default in the library, this will always happen.
For most users setting this up, will decrease performance.
 
You can always do:
var NATS = require('nats');
NATS.MAX_CONTROL_LINE_SIZE = 1024; (or match server or any custom size you provide to the server.).

aricart:
@sgtram Ran a new set of tests and finding that the performance on node 9 is better. Going to push a new change bumping the max control line to 1024.

Issue #174, opened 2017-11-07
node-nats doesn't have timer for client pings to the server


aricart:
node-nats doesn't have a timer for client pings to the server resulting in very long timeout notifications. See #173.

PR #179, opened 2017-11-16
[FIX] #174 added timer operation, tests.

aricart:


lib/nats.js
```diff
@@ -547,6 +595,8 @@ Client.prototype.createConnection = function() {
     }
     // Create the stream
     this.stream = net.createConnection(this.url.port, this.url.hostname);
+    this.stream.id = nuid.next();
+    this.emit('created', this.stream.id);
```
kozlovic: Do you need those 2 lines? I don't see the stream's 'created' event being used in the new test.

lib/nats.js
```diff
@@ -999,7 +1052,8 @@ Client.prototype.processErr = function(s) {
     // except stale connection and permission errors
     var m = s ? s.toLowerCase() : '';
     if (m.indexOf(STALE_CONNECTION_ERR) !== -1) {
-        this.scheduleReconnect();
+        this.closeStream();
+        // this.scheduleReconnect();
```
kozlovic: Remove this and maybe instead add comment above to say that `closeStream()` will trigger reconnect if allowed.

test/support/mock_server.js
```diff
@@ -0,0 +1,149 @@
...
+    socket.write("PING\r\n");
+}
+
+exports.sendPing = sendPing;
+
+function sendPong(socket) {
+    socket.write("PONG\r\n");
+}
+
+exports.sendPong = sendPong;
```
kozlovic: nit: do you need new line here?

aricart: I added it as I showed you. For some reason, they are getting stripped out of the diff.

PR #180, opened 2017-11-29
docs : list init options, and their default values

bguiz:
What:

- [x] find what all the options are
- [x] list them together with their default values

Why:

- Currently, some of these options are shown in examples in the README,
  however, there isn't a central place which lists them all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,34 @@ nc.on('close', function() {
 
 See examples and benchmarks for more information.
 
+## Options
+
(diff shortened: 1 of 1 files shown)


aricart:
I think the correct way to do this is to add a function that generates the defaults and then dumps them. This way we can expose the defaults without making them go out of date.

aricart:
The README.md file has been updated with all properties, default values, and documentation.

PR #187, opened 2018-02-12
NodeJS types changed TlsOptions to TlsServerOptions

Qwerios:
Updated for refactored NodeJS typings

Source: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23393/commits/e00c71172a6d67a538b27e7cfcdd1370fabdf959

This broke literally 2 days ago :( I have other TypeScript packages (like pg) that already updated to tls.TlsServerOptions so I couldn't get away with pinning @types/node-js.

Kind of a nasty gotcha because other people might have older dependencies and can't upgrade. In a perfect world we could all be up to date all the time.

If you do have the option to pin: "@types/node": "9.4.4"

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -61,7 +61,7 @@ export interface ClientOpts {
 	servers?: Array<string>,
 	noRandomize?: boolean,
 	encoding?: BufferEncoding,
-	tls?: boolean | tls.TlsOptions,
+	tls?: boolean | tls.TlsServerOptions,
 	name?: string,
 	yieldTime?: number,
 	waitOnFirstConnect?: boolean,


aricart:
@Qwerios I don't have the option to clamp the version of typescript that clients use, as node-nats is not written in typescript, it simply exports a typescript definition. So this change will trigger the same issue in the other direction. However, I don't see another workable strategy unless I separate the typescript binding from the node-nats library. I will merge for now

Qwerios:
Thanks for merging. I posted the pinning thing more for people coming here looking for a temporary solution.

veeramarni:
When would be the next release? I'm looking for this fix so my builds don't fail.

aricart:
@veeramarni v.0.7.25 contains this fix and it is available in npm.

Qwerios:
@aricart  I am losing the will to live here....they have once again undone the rename

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23615

Would you like me to make another pull request?

aricart:
I got it. - I thought it weird that they would rename a type.

Qwerios:
Same here but at least they are now discussing how to prevent a snafu like this in the future

Issue #192, opened 2018-02-13
nats process does not stop from version 0.7.25


imrefazekas:
I updated to version 0.7.25 and  after nats.close() the node process is still running.

Running it with 0.7.24 works fine and node process stops properly.

imrefazekas:
This code does not stop with 0.7.25 but stops with 0.7.24.
I'm using node 9.4.0
```
let NATS = require('nats')

try {
	let nats = NATS.connect( {
		'url': 'nats://localhost:4222'
	} )

	nats.on('error', function (err) {
		console.log(err)
	})

	nats.on('connect', function (nc) {
		console.log('connected')
	})

	nats.on('disconnect', function () {
		console.log('disconnect')
	})

	nats.on('reconnecting', function () {
		console.log('reconnecting')
	})

	nats.on('reconnect', function (nc) {
		console.log('reconnect')
	})

	nats.on('close', function () {
		console.log('close')
	})

	setTimeout( () => {
		nats.close()
	}, 5000 )
} catch (err) {
	console.error( err )
}
```

aricart:
@imrefazekas I have verified the issue. The heartbeat code is keeping it from exiting for a long time.

PR #193, opened 2018-02-14
[FIX] #192 - the heartbeat timer prevents nats.close() from existing Close now removes the timer if it was scheduled.

aricart:


derekcollison:
+1 to a test.

aricart:
"If we test it we may find other bugs"
Interestingly we don't seem to emit 'close'... This will require a few more calories/review.

derekcollison:
We had tests that checked for close emit at some point.

aricart:
@derekcollison the `close` event is propagated when the client disconnects from a server and has exhausted its retries). When initiated by a client calling `close()`, no notification is given because the code itself called it. I propose that we add a `closed` event that is used in a test. Alternatively, we can launch a subprocess and verify it exits.

derekcollison:
We should do a test for exit like you suggest. Not sure on closed event when application calls close.

Issue #151, opened 2017-06-05
Requests are too protocol chatty, should be optimized


petemiron:
re: https://github.com/nats-io/go-nats/issues/294

Must retain request and requestOne semantics, while minimizing network calls to subscribe to a a new inbox channel.

aricart:
Final fix is in #196

PR #195, opened 2018-02-27
New request reply

aricart:
- Implements 'new' multiplexed request reply via 'requestOne', where a single multiplexed subscription is reused for all requests. Old requestOne behaviour can be used by setting the option `useOldRequestStyle` option to true.

- Fixes argument shifting in oldRequestOne.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1443,12 +1443,19 @@ Client.prototype.initMuxRequestDetails = function(callback) {
 Client.prototype.cancelMuxRequest = function(token) {
 
     // if the token is a number, we have a fake sid, find the request
-    if(typeof token === 'number') {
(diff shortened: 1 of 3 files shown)


lib/nats.js
```diff
@@ -1334,12 +1344,149 @@ Client.prototype.request = function(subject, opt_msg, opt_options, callback) {
  * @param {String} opt_msg
  * @param {Object} opt_options
  * @param {Function} callback - can be called with message or NatsError if the request timed out.
```
kozlovic: Missing timeout from list of parameters here (may want to indicate that it is required)

aricart:
This PR addresses #151.

derekcollison:
LGTM

PR #199, opened 2018-03-01
Fix https://github.com/nats-io/node-nats-streaming/issues/58

aricart:
[FIX] Handling of calling `nats.timeout()` with an SID generated by t…he new request/reply. This fixes https://github.com/nats-io/node-nats-streaming/issues/58.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1274,6 +1274,9 @@ Client.prototype.unsubscribe = function(sid, opt_max) {
 /**
  * Set a timeout on a subscription. The subscription is cancelled if the
  * expected number of messages is reached or the timeout is reached.
+ * If this function is called with an SID generated  multiplexed
+ * request call, the original timeout handler associated with the multiplexed
+ * request is replaced with the one provided to this function.
  *
  * @param {Mixed} sid
  * @param {Number} timeout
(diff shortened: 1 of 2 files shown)


lib/nats.js
```diff
@@ -1274,6 +1274,9 @@ Client.prototype.unsubscribe = function(sid, opt_max) {
 /**
  * Set a timeout on a subscription. The subscription is cancelled if the
  * expected number of messages is reached or the timeout is reached.
+ * If this function is called with an SID generated  multiplexed
```
kozlovic: `a` and extra space between `generated` and `mutiplexed`

lib/nats.js
```diff
@@ -1438,13 +1453,13 @@ Client.prototype.createResponseMux = function() {
         var client = this;
         var inbox = createInbox();
         var ginbox = inbox + ".*";
-        var sid = this.subscribe(ginbox, function(msg, reply, subject) {
+        var sid = this.subscribe(ginbox, {}, function(msg, reply, subject) {
```
kozlovic: why do we pass empty options here? Is it that subscribe() is broken otherwise?

aricart:
Reworked the doc for timeout. Also the extra sub options was prompted by the IDE showing that an extra argument was needed. This is due to an incorrect definition of subscribe in the index.d.ts file which was contributed. The argument is optional.

PR #206, opened 2018-03-09
Options documentation

aricart:
Added documentation listing options and default values.


README.md
```diff
@@ -261,8 +261,40 @@ nc.on('close', function() {
...
+|--------                |---------                                     |---------                  |------------
+| `encoding`             |                                              | `"utf8"`                  | Encoding specified by the client to encode/decode data
+| `json`                 |                                              | `false`                   | If true, message payloads are converted to/from JSON
+| `maxPingOut`           |                                              | `2`                       | Max number of pings the client will allow unanswered before rasing a stale connection error
+| `maxReconnectAttempts` |                                              | `10`                      | Sets the maximun number of reconnect attempts. The value of `-1` specifies no limit
+| `name`                 | `client`                                     |                           | Optional client name
+| `noRandomize`          | `dontRandomize`, `NoRandomize`               | `false`                   | If set, the order of discovered servers in not randomized
```
kozlovic: It actually applies to the list of servers that you provide as an options to the connect (at least in other languages you can have an array of server urls or a list of servers as the connect url). The discovered server list is already coming to the client in no particular order so randomization should not apply there (recent PR [here](https://github.com/nats-io/go-nats/pull/344) removes the attempt to randomize the incoming list of discovered servers for the reason stated above).

PR #213, opened 2018-03-13
Server pool update

aricart:
server pool update enhancements.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -298,6 +298,16 @@ Client.prototype.parseOptions = function(opts) {
         options.servers.forEach(function(server) {
             client.servers.push(new Server(url.parse(server)));
         });
+      // Randomize if needed
+      if (options.noRandomize !== true) {
+        shuffle(client.servers);
+      }
+
+      // if they gave an URL we should add it if different
+      if(options.url !== undefined && client.servers.indexOf(options.url) === -1) {
+        //make url first element so it is attempted first
+        client.servers.unshift(new Server(url.parse(options.url)));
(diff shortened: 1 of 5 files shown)


lib/nats.js
```diff
@@ -1003,6 +992,57 @@ Client.prototype.processInbound = function() {
     }
 };
 
+/**
+ * Process server updates in info object
+ * @api internal
+ */
+Client.prototype.processServerUpdate = function() {
+  var client = this;
+  if (client.info.connect_urls && client.info.connect_urls.length > 0) {
```
kozlovic: As it happens, the new server will now send its own URL(s) as part of that INFO protocol, otherwise this `if` would have prevented from updating the server pool if all servers disconnected from the cluster. (note that older servers are not sending updates when servers go away, but would also not send their own urls).

lib/nats.js
```diff
@@ -1003,6 +992,57 @@ Client.prototype.processInbound = function() {
...
+    // perform the deletion
+    toDelete.reverse();
+    toDelete.forEach(function(index) {
+      client.servers.splice(index, 1);
+    });
+
+    // remaining servers are new
+    var newURLs = [];
+    for(var k in tmp) {
+      if(tmp.hasOwnProperty(k)) {
```
kozlovic: What does that mean?
At this point, you should add to server pool all the URLs that are in `tmp`, and detect the ones that were never seen before, so that we pass those to the `serverDiscovered` callback.
I can't say if this is what this code is doing here...

PR #225, opened 2018-05-02
Added full JSON type support in messages

GeorgeSapkin:
[`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) supports types other than an object and an array (e.g. string, number, etc). I've removed the `typeof` check to prevent `BAD_JSON` errors.


GeorgeSapkin:
Tests seem to be very flaky in Travis and randomly failing on different runs. Not sure if I can do anything about this in the scope of this PR.

derekcollison:
Be good to have more negative tests. Also, if I pass a string and its already properly or improperly formatted as json, what should the tests do when the JSON option is set?

aricart:
We need the following tests:

{}, [], "", number, true, false, null.
Also if I recall correctly setting JSON option to be true, makes payload parameters non-optional, so we need tests that don't provide them. Or handle them accordingly.

aricart:
I made the tests for `[], "", number, true, false, null` in go-nats. It works!

aricart:
@GeorgeSapkin could you add test for RequestReply. 

```
    function testRequest(input, useOldRequestStyle) {
        return function(done) {
            var opts = {
                json: true,
                port: PORT,
            };
            if(useOldRequestStyle) {
                opts.useOldRequestStyle = true;
            }

            var nc = NATS.connect(opts);
            nc.subscribe('request', {max: 1}, function(msg, reply) {
                nc.publish(reply, msg);
            });

            nc.request('request', input, function(msg) {
                if(msg instanceof Array) {
                    msg.should.deepEqual(input);
                } else {
                    (input === msg).should.be.true;
                }
                done();
            });
        };
    }

    it('should req/rep with null message', testRequest(null));
    it('should req/rep with null message oldrr', testRequest(null, true));
    it('should req/rep with integer', testRequest(122));
    it('should req/rep with integer oldrr', testRequest(122, true));
    it('should req/rep with empty string', testRequest(""));
    it('should req/rep with empty string oldrr', testRequest("", true));
    it('should req/rep with string', testRequest("abc"));
    it('should req/rep with string oldrr', testRequest("abc", true));
    it('should req/rep with true', testRequest(true));
    it('should req/rep with true oldrr', testRequest(true, true));
    it('should req/rep with empty array', testRequest([]));
    it('should req/rep with empty array oldrr', testRequest([], true));
    it('should req/rep with array', testRequest([1, 2, 'foo']));
    it('should req/rep with array oldrr', testRequest([1, 2, 'foo'], true));
```

lib/nats.js
```diff
@@ -1166,7 +1166,7 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
         opt_callback = subject;
         subject = undefined;
     }
-    if (!msg) {
+    if (!msg && !this.options.json) {
         msg = EMPTY;
```
aricart: if msg is undefined, at the very least the RR test I put will fail.

lib/nats.js
```diff
@@ -1166,9 +1166,16 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
...
         subject = undefined;
     }
-    if (!msg) {
-        msg = EMPTY;
+
+    if (!this.options.json) {
+        if (!msg) {
+            msg = EMPTY;
+        }
+    } else if (msg === undefined) {
```
aricart: I think the correct thing to do here is to make msg `null` - On other clients where the value is required, requests/publishes will have to provide null/nil, which will serialize as `null` in the JSON. We need to treat the default message to be `null` in the case of json.

The publish and request need to add a test where they elide the payload, and the subscribed value is expected to have been a serialized `null` json.

lib/nats.js
```diff
@@ -1166,9 +1166,16 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
         opt_callback = subject;
         subject = undefined;
     }
-    if (!msg) {
-        msg = EMPTY;
+
+    if (!this.options.json) {
+        if (!msg) {
```
aricart: The test here will need to be more careful - as the provided values could be `false`, `""` (empty string), `null`, `undefined`, and only in the case of undefined should this value be touched to be `null`

aricart:
when travis is slow and tests that require multiple servers start, the whackamole restart builds happens.  Hang tight.

.eslintrc.js
```diff
@@ -1,5 +1,7 @@
 module.exports = {
     "env": {
+        "es6": true,
```
aricart: es6 shouldn't be part of this.

GeorgeSapkin: I'll remove it. However, you are targeting Node 6.x+ according to [README](https://github.com/nats-io/node-nats#supported-node-versions). Without it linter complains about every ES6 feature that's part of Node 6.x. Should this be in a separate PR then?

test/json.js
```diff
@@ -14,79 +14,140 @@
...
         server = nsc.start_server(PORT, done);
     });
 
     // Shutdown our server
-    after(function(done) {
+    after((done) => {
         nsc.stop_server(server, done);
     });
 
+    function testPubSub({ input, expected }) {
```
aricart: While we are not testing on node 4.5, our source is compatible and works correctly on node 4.5+. The object destructuring is wicked cool, but not supported in nodejs4.

test/json.js
```diff
@@ -14,79 +14,140 @@
...
+                nc.close();
+
+                done();
+            });
+
+            nc.publish('pubsub', input);
+        };
+    }
+
+    function testReqRep({ input, expected, useOldRequestStyle }) {
```
aricart: While we are not testing on node 4.5, our source is compatible and works correctly on node 4.5+. The object destructuring is wicked cool, but not supported in nodejs4.

lib/nats.js
```diff
@@ -1166,9 +1166,16 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
         opt_callback = subject;
         subject = undefined;
     }
-    if (!msg) {
-        msg = EMPTY;
+
```
aricart: this would be simpler:

```
    if(!this.options.json) {
        msg = !msg ? EMPTY : msg;
    } else {
        // undefined is not a valid JSON-serializable value, but null is
        msg = msg === undefined ? null : msg;
    }
```

aricart:
one nitpick and one compatibility issue. Otherwise LGTM. If you address, I will merge it. Thanks for your contribution @GeorgeSapkin

test/json.js
GeorgeSapkin: Maybe make it clear in the README then. Right now it says you're only targeting releases that are in maintenance or newer. Which is 6.x and newer.

I'll remove the destructuring.

GeorgeSapkin:
I've removed destructuring. Is there anything else I'm missing?

aricart:
It's great. Thanks @GeorgeSapkin!

PR #227, opened 2018-06-15
version bump

aricart:


lib/nats.js
```diff
@@ -863,7 +863,7 @@ Client.prototype.processInbound = function() {
                         cb();
                     } // FIXME: Should we check for exceptions?
                 } else if ((m = PING.exec(buf)) !== null) {
-                    client.sendCommand(PONG_RESPONSE);
+                    client.sendCommand(PONG_RESPONSE);p
```
cjbhaines: @aricart did you mean to add this p?

aricart: that was an error.  I pulled the request, out since the version was already bumped for 1.0.

cjbhaines: Ah right I see, cool. Thanks for doing the release!

PR #231, opened 2018-07-20
gnatsd 1.2.0 requires the client to check info values before initiating a TLS connection.

aricart:
[FIX] Attempting to connect via TLS on a server that doesn't have TLS available (as reported by the INFO), now results in a different error. Fixed a check where we verify if the info tls_required is set.

Bumped travis gnatsd to 1.2.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ git:
   depth: false
 
 env:
-  - NATS_VERSION=v1.0.6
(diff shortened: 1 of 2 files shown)


aricart:
IMHO all values in the info should be provided - that way there's no games on what the value could be. Depending on the parser it will have very different results. In go mapping to the struct will give a default value so it works. In other languages, it is now null or undefined. 
@derekcollison

Issue #233, opened 2018-09-13
Message stream broken on error in callback


valichek:
Exception thrown in callback will break all the following messages. The reason is that execution will stop at this line: https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L971, `client.pstate` will never changed and parsing of `client.inbound` will break.

StarpTech:
That's the typical behavior of synchronous errors in Node.js you have to `try/catch` it.

valichek:
@StarpTech sure it's typical, but the library is async, so I would try call callbacks async as well, maybe `setImmediate()` could be used here https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L1093 to fix this problem.

StarpTech:
I think I don't understand your case. Could you make a repro? Snippet?

valichek:
@StarpTech smth like
```
nats.subscribe('help', function(request, replyTo) {
  throw new Error('this error will break all subsequent received messages')
});
```

StarpTech:
As I expected this can't be fixed. You have to catch the error. It's by design that the lib doesn't catch the error for you.

valichek:
well, if using `setImmediate()` is not somehow a solution here, this should be documented, but it's really weird that one exception can break everything and be even harmful for the data

StarpTech:
What do you expect from `setImmediate()`? It will postpone the execution but will still throw the error asynchronously. It's your responsibility to write save code.

```js
nats.subscribe('help', function(request, replyTo) {
  try {
  	opThatCanThrow()
  } catch(err) {
	logStuff(err)
  }
});
```

StarpTech:
@valichek For better understanding https://hemerajs.github.io/hemera/docs/error-handling.html

valichek:
@StarpTech the problem is not to catch error, problem is that not catched error will influence to all the received data and subscriptions. What if I don't want to catch it? This can be the case as well. This error will produce errors in other callbacks and it will be really hard to track the reason.

kozlovic:
I don't know enough Node.js to have a valuable input, but unless in Node if the library were to catch the error it would prevent the user to catch it, then the library should catch any error to prevent misbehaving user callback to cause the library to fail.

aricart:
[...]

PR #234, opened 2018-09-13
[FIX] #233 - if the client handler throws an exception, the library is possibly left in an inconsistent state.

aricart:


StarpTech:
How do you want to fix it with `set Immediate` I don't get it. I also don't like this solution because as @valichek said the client should not responsible for user space errors. In worst case the application will exit and restarted. Do I miss something or isn't this the case?

valichek:
@StarpTech smth like 
```
// msg is already local variable, so we set only these for closure
var reply = this.payload.reply;
var subj = this.payload.subj;
var sid = this.payload.sid;
var cb = function() {
    sub.callback(msg, reply, subj, sid);
}
setImmediate(cb);
```

StarpTech:
@valichek as I asked before what do you expect? The process will still crash. Do you want to save inbound messages?

valichek:
@StarpTech the first - is that this will result inbound buffer parsing to finish with consistent state and thrown error will not affect other messages - so it will not stop at this line https://github.com/nats-io/node-nats/blob/master/lib/nats.js#L971, another thing is that callbacks will be executed in async - not one by one from single flow like now.

valichek:
@StarpTech  it's important to keep inbound buffer consistent, because next call of `processInbound` can be executed before crash

StarpTech:
@valichek thanks for clarification now it's clear for me.

> it's important to keep inbound buffer consistent, because next call of processInbound can be executed before crash

What's the benefit of this? Even though we delay the execution to the next tick the first uncaught error will crash the process and all other callback have no chance to complete. You left your application in a broken state and you shouldn't expect nothing.

valichek:
@StarpTech exception does not always result in crash, you can handle exceptions at very high level and decide how to proceed, for more example you can check https://stackoverflow.com/questions/7310521/node-js-best-practice-exception-handling

StarpTech:
@valichek but this is not what your issue is about. If use `try/catch` you can always handle it anyway. Your proposal was to add `setImmediate` which IMO doesn't help here.

valichek:
Btw: that browser js app never crashes, so it will definitely go to inconsistent state without any chance to proceed normally

StarpTech:
This library is for server-side only.

valichek:
@StarpTech actually, when you handle errors on higher levels of app and decide not to crash, `setImmediate()` will help, because exception will effect only to the current callback

valichek:
@StarpTech what about https://nodejs.org/api/process.html#process_event_uncaughtexception ?

StarpTech:
This is an anti-pattern. Read the `warning` description 😄

valichek:
@StarpTech warning say nothing about being anti-pattern, sure one should take care about what he is doing

StarpTech:
Exactly and in this case it's an antipattern because you wanna use it as a global catch handler not to gracefully shutdown your application.

aricart:
> re-emitting caught error seems not good solution to me, because error doesn't belong to the library, but to the user code, using `setImmediate()` would be better solution, even if you loose a part of stack trace - it will be just nats related lines, btw `setImmediate()` is already used, so in some cases stack trace will not contain even a caller of `processInbound()`

`setImmediate` doesn't work. It happens when the event loop returns, which is when the inbound buffer is drained.

StarpTech:
@aricart please consider this change as a breaking change because with this change user errors are swallowed when someone registered the `error` handler to listens only for NATS errors 😟

aricart:
@StarpTech there's no api guarantee that the error you get from the the `error` handler is a NATSError. In the case where we raise the error, they will always be. This is already an user exception that is not handled.

PR #235, opened 2018-11-01
Add some examples

minhnt95:
Demo for some messaging mechanisms of Nats.io:
- Simple pub/sub
- Request to many / reply
- Request to only one / reply
- Queue sharing


aricart:
@minhnt95 Thanks for the contribution. These all overlap with our published samples, on https://www.nats.io/documentation which have been carefully selected and developed cross-language.

aricart:
@minhnt95 could you take a look at https://github.com/nats-io/node-nats-examples and see if your samples can be turned into tooling that can be built and tested on the examples site?

minhnt95:
@aricart thanks for your replies

PR #239, opened 2018-12-08
Added support for nkeys, user jwts, and signing.

derekcollison:
Added nkey, jwt, and sig support.
Added auto-upgrade to tls.
Fixed new style request emit unsubscribe event (for new node-req example)
Added simplified connect with second parameter option creators. We can use the nats.cred(filepath) here. e.g. nats.connect(url, opts)

Signed-off-by: Derek Collison <derek@nats.io>


aricart:
LGTM - Not sure why we are emitting unsubscribe on a mux subscription for request-reply but sounds like it could be useful.

PR #247, opened 2019-02-25
Update TypeScript definition of messages. Fixes #246

### LINKED ISSUE #246: Typings don't support publishing objects
When automatic JSON conversion is enabled, one would want to use the `publish` function like
```
const mymessage = { a: 1, b: "2"};
nats.publish("mysubject", mymessage);
```
However, the current Typescript signatures of the publish function do not support this:
```
publish(callback: Function):void;
publish(subject: string, callback: Function):void;
publish(subject: string, msg: string | Buffer, callback: Function):void;
publish(subject: string, msg?: string | Buffer, reply?: string, callback?: Function):void;
```
Instead, they should be
```
publish(callback: Function):void;
publish(subject: string, callback: Function):void;
publish(subject: string, msg: string | Buffer | object, callback: Function):void;
publish(subject: string, msg?: string | Buffer | object, reply?: string, callback?: Function):void;
```

jorispz:


aricart:
I believe the proper definitions for publish are:

```
	publish(subject: string, callback: Function):void;
	publish(subject: string, msg: any, callback: Function):void;
	publish(subject: string, msg: any, reply: string, callback?: Function):void;
```

For request, they are somewhat ill-formed, as the callback is always required:
```
	request(subject: string, callback: Function): number;
	request(subject: string, msg: any, options: SubscribeOptions, callback: Function): number;
```

derekcollison:
Can we have request do a more async/wait or promise syntax vs callback?

aricart:
@derekcollison - yes by using ts-nats. There's no point in starting to mix the event library with the promise/async library.

jorispz:
Thanks for the feedback, I updated the TS definitions to your proposal.

For my personal project I switched over to ts-nats, because the async/await syntax suits my problem much better. Thanks for working on these libraries!

aricart:
@jorispz thanks for the contribution - glad you are liking ts-nats.

PR #254, opened 2019-03-25
ES6 linting

aricart:
- changed var statements to const/let
- added dev dependency on @types/node to remove warnings from index.d.ts
- migrated istanbul to nyc coverage
- added package lock

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2018 The NATS Authors
+ * Copyright 2013-2019 The NATS Authors
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -13,13 +13,14 @@
  * limitations under the License.
  */
 
-/* jslint node: true */
+/* eslint no-sync: 0 */
+
 'use strict';
 
 /**
  * Module Dependencies
  */
-var net = require('net'),
+const net = require('net'),
     tls = require('tls'),
     url = require('url'),
     util = require('util'),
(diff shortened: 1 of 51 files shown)


derekcollison:
What versions of Node.js does this cut off?

aricart:
@derekcollison This cuts 4 which died a long time ago - 6 is also about to die on 4/30.

derekcollison:
Agree, but wondering if we still have users on 4..

aricart:
if we do, they are on their own on many fronts. Also the release history has versions that accommodate them.

aricart:
@kozlovic - node-nats tests on travis have a 90% chance of failure always - they require babysitting. Imho most are invalid timeouts.

examples/notifications
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env node
+/*
+ * Copyright 2013-2018 The NATS Authors
```
derekcollison: 2019?

lib/nats.js
```diff
@@ -627,74 +625,73 @@ Client.prototype.scheduleHeartbeat = function() {
...
     if (undefined === stream) {
         return;
     }
 
-    stream.on('connect', function() {
-        if (client.pingTimer) {
-            clearTimeout(client.pingTimer);
-            delete client.pingTimer;
+    stream.on('connect', () => {
+        if (this.pingTimer) {
```
derekcollison: We sure this is safe? Maybe the new syntax makes this ok but I remember that `client=this` was a big deal back in the day.

aricart: Yes that is the whole thing with the arrow functions - the this there is the one in the outer block, which in this case is the client. It removes all sorts of ambiguities.

derekcollison: ok glad they fixed that about the language - lol. I am sure they did a long time ago.. I am slow.

examples/notifications
aricart: Yes need to update all the touched ones.

derekcollison: Fix all of them that you touched.

Issue #257, opened 2019-03-27
disconnect should only be notified if the current connection fails


aricart:
Currently, as the node-nats tries to establish a connection, the stream handler notifies of a disconnect even when the client was not able to connect.

Issue #256, opened 2019-03-26
node-nats can leave request timeouts active after closing


aricart:
node-nats is not cleaning up request timeouts.
On close, all request timeouts should be canceled:
```
Client.prototype.clearMuxTimeouts = function() {
    if(this.respmux) {
        for (var p in this.respmux.requestMap) {
            if (this.respmux.requestMap.hasOwnProperty(p)) {
                this.cancelMuxRequest(p)
            }
        }
    }
}
```

aricart:
Or for that matter subscriptions that specify a timeout.

PR #258, opened 2019-03-27
Timer leaks/Multiple disconnect notifications

### LINKED ISSUE #255: node-nats leaves ping timer out when disconnected triggers a close
stream.on('close performs the logic on whether to fire the close event, but it doesn't clean ping timeout.

``` 
stream.on('close', function(hadError) {
// needs the following added:
        if (client.pingTimer) {
            clearTimeout(client.pingTimer);
            delete client.pingTimer;
        }
```

aricart:
FIX #255, #256, #257.

Currently, when the client is closed, only the ping timer is cleaned up. Other timers such as timeouts on a subscription or a request would remain live.
Fixed so that if the client fails to reconnect and issues a 'close' from the stream, that timeout resources are also cleaned up.

Disconnect notifications are raised by the stream handler whenever close is called on the stream, this is incorrect because the close handler is called even if a connection on the current socket was not established. This makes downstream clients (such as node-nats-streaming) have false notifications that complicate the logic of tests and other things. With this change a 'disconnect' is only notified if the current socket was actually active.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -651,7 +651,9 @@ Client.prototype.setupHandlers = function() {
 
     stream.on('close', (hadError) => {
         this.closeStream();
-        this.emit('disconnect');
+        if(stream.bytesRead > 0) {
+            this.emit('disconnect');
+        }
         if (this.closed === true ||
             this.options.reconnect === false ||
             ((this.reconnects >= this.options.maxReconnectAttempts) && this.options.maxReconnectAttempts !== -1)) {
@@ -670,7 +672,7 @@ Client.prototype.setupHandlers = function() {
 
         // if the current server did not connect at all, and we in
         // general have not connected to any server, remove it from
-        // this list. Unless overidden
+        // this list. Unless overridden
         if (this.wasConnected === false && this.currentServer.didConnect === false) {
             // We can override this behavior with waitOnFirstConnect, which will
             // treat it like a reconnect scenario.
@@ -874,6 +876,7 @@ Client.prototype.cleanupTimers = function() {
                 const sub = this.subs[p];
                 if(sub.timeout) {
                     clearTimeout(sub.timeout);
+                    delete sub.timeout;
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


lib/nats.js
```diff
@@ -831,10 +843,7 @@ Client.prototype.initState = function() {
  * @api public
  */
 Client.prototype.close = function() {
-    if (this.pingTimer) {
-        clearTimeout(this.pingTimer);
-        delete this.pingTimer;
-    }
+    this.cleanupTimers();
```
aricart: this is not needed

kozlovic: why?

aricart: close on the stream will do it.

kozlovic: Ok, we will still have all the subscriptions by that time, right? That is, there is no place where we would remove them by the time we get to stream.close and therefore are unable to cancel subscriptions' timer?

aricart: actually - my bad the cleanup is required as closing the stream from `close` will not emit the 'close' event on the stream.

lib/nats.js
```diff
@@ -588,7 +588,19 @@ Client.prototype.connectCB = function() {
     this.flushPending();
 };
 
+/**
+ * @api private
+ */
+Client.prototype.cancelHeartbeat = function() {
+    if (this.pingTimer) {
```
derekcollison: Should we not set pingTimer to null or undefined in case this gets called again?

lib/nats.js
```diff
@@ -846,6 +853,33 @@ Client.prototype.close = function() {
...
+ */
+Client.prototype.cleanupTimers = function() {
+    this.cancelHeartbeat();
+
+    if(this.respmux && this.respmux.requestMap) {
+        for (const p in this.respmux.requestMap) {
+            if (this.respmux.requestMap.hasOwnProperty(p)) {
+                this.cancelMuxRequest(p);
+            }
+        }
```
derekcollison: Again should we null or set to undefined in case of being called again?

PR #260, opened 2019-03-29
FIX #259 - implementation of noEcho was never merged into master.

### LINKED ISSUE #259: `noEcho` needs implementation

aricart:


lib/nats.js
```diff
@@ -118,6 +118,7 @@ const VERSION = '1.2.4',
     NO_USER_JWT_IN_CREDS_MSG = 'Can not locate user jwt in credentials.',
     BAD_OPTIONS = 'BAD_OPTIONS',
     BAD_OPTIONS_MSG = 'Options should be an object as second paramater.',
+    NO_ECHO_NOT_SUPPORTED = 'NO_ECHO_NOT_SUPPORTED',
```
kozlovic: Should you use a `_MSG` for the error text?

lib/nats.js
```diff
@@ -430,6 +433,15 @@ Client.prototype.selectServer = function() {
     this.servers.push(server);
 };
 
+Client.prototype.checkNoEchoMismatch = function() {
+    if ((this.info.proto === undefined || this.info.proto < 1) && this.options.noEcho) {
+        this.emit('error', new NatsError(NO_ECHO_NOT_SUPPORTED, NO_ECHO_NOT_SUPPORTED));
```
kozlovic: Should you pass the text and not the code twice?

lib/nats.js
```diff
@@ -749,6 +761,9 @@ Client.prototype.sendConnect = function() {
     if (this.options.nkey !== undefined) {
         cs.nkey = this.options.nkey;
     }
+    if (this.options.noEcho) {
+        cs.echo = false;
```
kozlovic: Humm... `cs.echo` should be initialized to `true` and set to `false` if `noEcho` option is set (as you do). So without `noEcho: true` your test that check that by default a connection receives its own message should fail?

lib/nats.js
aricart: yeap.

lib/nats.js
aricart: interesting - not sure why the test doesn't fail.

aricart: perhaps on the server unless the option is returned, it doesn't do anything

aricart: Regardless I cannot set it unless they provide the option, because if the server doesn't support the option, it will fail.

kozlovic: Ok. After discussion, it works because the server defaults the "echo" boolean to true in the structure that is unmarshal'ed when processing the client CONNECT. So that old clients that don't set this work ok (that is with "echo" semantics).
IMO, you should set it to `true` by default and to `false` if user sets NoEcho, but either way works. (I believe that if you set to `true` and connect to a server that does not know about the `echo` field in CONNECT, nothing bad will happen).

aricart: Seems to work correctly under 1.1.0 - but keeping it as is for right now.

PR #262, opened 2019-04-08
FIX #261 TypeScript bindings

### LINKED ISSUE #261: Typings don't support publish(subject, message) call
Basic usage example is not supported by typings
```javascript
nats.publish('foo', 'Hello World!');
```

In the commit fixing #246, reply was made mandatory https://github.com/nats-io/node-nats/commit/70a055d993b2c0e2bd052cb7e3b732143871f138#diff-b52768974e6bc0faccb7d4b75b162c99R117 and thus it can't be applied to the above call

aricart:
Contributed bindings were not exactly right. Fixed the typescript bindings to show the actual arguments that are required. Optional signature in typescript makes a previous argument required, but the library actually does argument shifting so the exploded signatures are the correct representation.

- removed a bad publish signature without a subject, but with callback
- clarified the publish with explicit arguments - really only optional argument given the three forms is the optional callback. Note the library does argument shifting so the exact forms are the expected.
- fixed requestOne signatures, to reflect that subject, timeout and callback are required.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,7 +40,7 @@ export const STALE_CONNECTION_ERR: string;
 
 /**
  * Create a properly formatted inbox subject.
-*/
+ */
 export function createInbox(): string;
 
 /**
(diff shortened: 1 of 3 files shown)


aricart:
Not really - let me see if we can do a local build of the bundle and then run the ts compiler on it.

StarpTech:
@aricart you could use https://github.com/Microsoft/dtslint or https://github.com/SamVerschueren/tsd

aricart:
@StarpTech thanks Dustin for the suggestions. The issue is not whether we wrote valid type definitions but rather that they didn't change in a way that they create 'fake' errors.  I changed the definition of some of the methods to address the notion that some values are optional. Unfortunately, node-nats does parameter reassignment so it is possible for different usages to be completely valid. To make it worse in some cases some optional parameters are followed by required ones, which as you know is illegal typescript. I think we got them right this go around, and really the right thing to do is to migrate to ts-nats where the declarations are part of the code rather than a grafted interface.

Issue #265, opened 2019-04-17
port tokenHandler callback from go-nats


andriy-bulynko:
In some use cases auth tokens might have expiry (like JWTs). In those cases, it would be nice to have a mechanism to ask for a new token before using it for authentication. `go-nats` already had that functionality added. It would be great to make `node-nats` feature compatible.

aricart:
@andriy-bulynko thanks for the PR - one issue I see is we don't use Promises anywhere on node-nats. Also, if you look we seem to have an ever-growing number of options for providing authentication - with the upcoming 2.0 nkeys, jwts, credential files, etc, we have a lot of handlers. I am starting to think that really what we should do is provide a connection option that takes a function and returns a connection option. Then this way all options could be dynamically resolved at the time of the connect while simplifying the API for providing these.

aricart:
@derekcollison wdyt?

derekcollison:
We need to look at more proactive re-authentication.

The Go client will read in the credentials files (new NATS 2.0 style security) such that the connection can be reestablished with the new credentials just by updating the file, but the connection will still be dropped by the server. We will do this work post 2.0.

aricart:
@derekcollison this is actually already implemented for the new authentication methods - jwt and creds, etc auto re-request the authentication configuration before connect.
A more general solution could be to allow clients to provide a function that return an auth options object that is refreshed for every connection.

derekcollison:
@aricart yes I remember, I wrote it ;)

What I mean is we are not proactive about expirations.

aricart:
lol

andriy-bulynko:
@aricart: Some callback before the connection to refresh the credentials would be extremely useful. Also, where could I read up on the upcoming planned JWT support you alluded to? Currently, we are already using `tokens` to pass JWTs, which is why we need a mechanism to refresh them.. Thanks!

aricart:
@andriy-bulynko the changes are on master and currently in production on our ngs service (see https://www.synadia.com/ngs).

You can read about some of the underlying tech in:
https://github.com/nats-io/jwt
https://github.com/nats-io/nkeys 
https://github.com/nats-io/nsc 

The nkey/jwt authentication effectively supersedes all the other mechanisms.
Docs and visibility of the functionality will show up in the next few days as we release for 2.0. Stay tuned!

PR #266, opened 2019-04-17
Port tokenHandler mechanism from go-nats (option processInbound())

andriy-bulynko:
Fixes: #265 

# Description

I wasn't sure of the best place to plug this in. This PR introduces the code directly in `processInbound()`, which is very close to how it's done in `go-nats`.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -79,6 +79,9 @@ const VERSION = '1.2.10',
     // Errors
     BAD_AUTHENTICATION = 'BAD_AUTHENTICATION',
     BAD_AUTHENTICATION_MSG = 'User and Token can not both be provided',
+    BAD_AUTHENTICATION_TH_NOT_FUNC_MSG = 'tokenHandler must be a function returning accepting a callback',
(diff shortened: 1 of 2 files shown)


aricart:
@andriy-bulynko I have the same comments as the other PR. We don't use Promises on node-nats.

andriy-bulynko:
@aricart: Ok, thanks. Just to be clear, is a callback approach like `tokenHandler(function(token, error){})` also off limits?

aricart:
@andriy-bulynko `tokenHandler(function(error, token)` would be fine. Just no Promise related code.

andriy-bulynko:
@aricart: Promises are out. Asynchronous callbacks are in.

andriy-bulynko:
Is this a problem with code or with build? Is there a way to kick the build to try again?

andriy-bulynko:
That's so weird. I just re-ran the tests on repeat a bunch of times and nothing failed.

andriy-bulynko:
There might be CI issues going on now...

aricart:
replaced with the sync version

PR #286, opened 2019-08-06
clean up argument handling in publish

basti1302:
This removes a few minor bugs from how the arguments in `publish` are handled:

1. `opt_callback` was called twice when `publish` was called without a subject but with a message and a reply (that is, `nc.publish(null, 'message', 'opt_reply', function(err) { ... })`) -- once with the `BAD_SUBJECT` error, once without (there was a missing return after the first call to the callback).
2. `opt_callback` was ignored and a type error was thrown synchronously when `publish` was called without a subject and without a reply (that is, `nc.publish(null, 'message', function(err) { ... })`)
3. A type error was thrown synchronously when passing four arguments but the fourth argument was not a function (because the code implicitly assumed that it is a valid callback).

In addition, I also cleaned up the tests around error handling a bit. Some of them were passing only accidentally (expecting a NatsError but only testing if any error is thrown, thereby masking the aforementioned TypeErrors). Also, two tests were testing multiple things at once -- I have split those tests so that each individual test takes care of one particular case.

I have run `npm test` successfully after my changes, all tests are passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,10 @@ Our support policy for Nodejs versions follows [Nodejs release support]( https:/
 We will support and build node-nats on even-numbered Nodejs versions that are current or in LTS.
 
 
+## Running The Tests
+
+To run the tests, you need to have a `nats-server` execubtable in your path. Refer to the [server installation guide](https://nats-io.github.io/docs/nats_server/installation.html#installing-via-a-package-manager) in the NATS.io documentation. With that in place, you can run `npm test` to run all tests.
+
 ## License
 
 Unless otherwise noted, the NATS source files are distributed under the Apache Version 2.0 license found in the LICENSE file.
--- a/index.d.ts
+++ b/index.d.ts
@@ -26,6 +26,7 @@ export const BAD_AUTHENTICATION: string;
 export const BAD_JSON: string;
 export const BAD_MSG: string;
 export const BAD_REPLY: string;
+export const BAD_CALLBACK: string;
 export const BAD_SUBJECT: string;
 export const CLIENT_CERT_REQ: string;
 export const CONN_CLOSED: string;
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -88,6 +88,8 @@ const VERSION = '1.3.0',
     BAD_MSG_MSG = 'Message can\'t be a function',
     BAD_REPLY = 'BAD_REPLY',
     BAD_REPLY_MSG = 'Reply can\'t be a function',
+    BAD_CALLBACK = 'BAD_CALLBACK',
+    BAD_CALLBACK_MSG = 'Callback needs to be a function',
     BAD_SUBJECT = 'BAD_SUBJECT',
     BAD_SUBJECT_MSG = 'Subject must be supplied',
     CLIENT_CERT_REQ = 'CLIENT_CERT_REQ',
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -441,6 +441,10 @@ Our support policy for Nodejs versions follows [Nodejs release support]( https:/
 We will support and build node-nats on even-numbered Nodejs versions that are current or in LTS.
 
 
+## Running The Tests
+
+To run the tests, you need to have a `nats-server` execubtable in your path. Refer to the [server installation guide](https://nats-io.github.io/docs/nats_server/installation.html#installing-via-a-package-manager) in the NATS.io documentation. With that in place, you can run `npm test` to run all tests.
```
basti1302: I took me a while to figure out that I need to install `nats-server` to get the tests going, so I thought I just document it for next clueless contributor coming along.

index.d.ts
```diff
@@ -26,6 +26,7 @@ export const BAD_AUTHENTICATION: string;
 export const BAD_JSON: string;
 export const BAD_MSG: string;
 export const BAD_REPLY: string;
+export const BAD_CALLBACK: string;
```
basti1302: Not sure if you are okay with introducing a new error code, but since I added a check for the type of the fourth argument, I tried to be consistent with the other validation checks.

lib/nats.js
```diff
@@ -1534,29 +1545,33 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
         msg = msg === undefined ? null : msg;
     }
 
+    if (opt_callback && typeof opt_callback !== 'function') {
+        throw (new NatsError(BAD_CALLBACK_MSG, BAD_CALLBACK));
+    }
```
basti1302: This guards against using the fourth argument as a callback when it is not a function.

lib/nats.js
```diff
@@ -1534,29 +1545,33 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
...
+
     if (!subject) {
         if (opt_callback) {
             opt_callback(new NatsError(BAD_SUBJECT_MSG, BAD_SUBJECT));
+            return;
```
basti1302: This missing return lead to `opt_callback` being called twice, once here with an error and one more time later, without an error.

lib/nats.js
```diff
@@ -1534,29 +1545,33 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
...
         } else {
             throw (new NatsError(BAD_SUBJECT_MSG, BAD_SUBJECT));
         }
     }
     if (typeof msg === 'function') {
-        if (opt_callback || opt_reply) {
+        if (opt_callback) {
```
basti1302: This check was broken: The code checks for the presence of opt_reply OR opt_callback, but then goes on to call `opt_callback` which of course fails when only `opt_reply` is present.

lib/nats.js
```diff
@@ -1534,29 +1545,33 @@ Client.prototype.publish = function(subject, msg, opt_reply, opt_callback) {
...
             opt_callback(new NatsError(BAD_MSG_MSG, BAD_MSG));
             return;
+        } else {
+            throw(new NatsError(BAD_MSG_MSG, BAD_MSG));
```
basti1302: Normalizing the arguments (checking if some optional arguments have been omitted and then using the last function argument as the callback) is now done at the top.

aricart:
@basti1302 I took a look at the PR, thank you for the contribution! - I think the parameter re-ordering to identify a callback need to be more generic and also needs to apply to any method that has optional arguments before a required callback. 

I have started a PR based on yours to try and make consistent the argument validation this initiates while ensuring that it is backward compatible. Keep tuned.

aricart:
PR #293 addresses the function argument issue for all api calls. - It is possible that that PR will be the basis for a 2.0 refactor of the API as it would be simpler to reorder some of the function parameters and introduce a parameter object for all optional arguments.

@basti1302  Thank you for your contribution, this sets in motion changes for a better experience.

PR #290, opened 2019-09-03
Fixing an issue with handling of the permission errors

### LINKED ISSUE #289: Incorrect handling of the permission errors
After the `permission_error` is handled, the `inbound` buffer is not cleared and all subsequent data appended to that buffer is ignored.

After @aricart 's explanation changing this to: after 1 `permission_error` is encountered, all subsequent buffer reads result in `permission_errors`.

andriy-bulynko:
Fixes #289

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1115,8 +1115,9 @@ Client.prototype.processInbound = function() {
                 } else if ((m = OK.exec(buf)) !== null) {
                     // Ignore for now..
                 } else if ((m = ERR.exec(buf)) !== null) {
-                    this.processErr(m[1]);
-                    return;
+                    if (this.processErr(m[1])) {
+                        return;
+                    }
                 } else if ((m = PONG.exec(buf)) !== null) {
                     this.pout = 0;
                     const cb = this.pongs && this.pongs.shift();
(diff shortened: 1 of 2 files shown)


lib/nats.js
```diff
@@ -1382,6 +1382,7 @@ Client.prototype.processErr = function(s) {
         this.closeStream();
     } else if (m.indexOf(PERMISSIONS_ERR) !== -1) {
         this.emit('permission_error', new NatsError(s, NATS_PROTOCOL_ERR));
+        this.inbound = null;
```
StarpTech: Why not calling ` this.closeStream();`?

StarpTech:
Please add a test.

lib/nats.js
andriy-bulynko: @StarpTech I don't think it's a terminal error. In our case this was happening on `publish()` calls to `subjects` that the user wasn't allowed to publish to. Other/allowed publish/subscribe activity could proceed as before. The `Golang` library does not treat this as a terminal error either.

StarpTech: Thanks for the insight.

andriy-bulynko:
@aricart: Are you saying that just the `permission_error` itself needs to be cleared from the `inbound` buffer? Not the whole buffer cleared? 
Please have a look at the test I added. Without removing the `permission_error` form that buffer, the library does not deliver "legal" messages after 1 `permission_error`. And `node-nats-streaming` library fails to process server ping responses...

aricart:
@andriy-bulynko I am saying that permission errors shouldn't strip the pending buffer. I think the bug that is in there, is that if you get a permission error, `processInbound()` is returning. And instead what it should do, is test if the error requires a disconnect - that is the only time it should return from processInbound. Otherwise it should just perform the notification as it does, and continue.

andriy-bulynko:
@aricart Ok, I see. I'll try to go along the path you are suggesting.. In the mean time, if you know off-hand what needs to be changed to fix this, please let me know and I'll patch this PR. Alternatively, if you'd like to take over the work to fix this bug, I'm okay with that too. Thanks in advance!

aricart:
I'll take the issue over. The fix is in the processInbound, and I rather own any changes there.

andriy-bulynko:
@aricart: Sure - that works! Btw, I made the change you suggested in this PR and it seems to have fixed the problem. All tests passed for me locally. However, I'm not 100% sure that those are all the changes that are needed.

aricart:
@andriy-bulynko - I didn't see that you made a PR, thanks for the contribution, it is almost the same as this one #291.

PR #291, opened 2019-09-04
FIX #289 -  permissions error aborted inbound processing

### LINKED ISSUE #289: Incorrect handling of the permission errors

aricart:
if a permissions error was encountered by the client, the connection remained open but processing of inbound was exited.

### DIFF
--- a/test/perms.js
+++ b/test/perms.js
@@ -39,7 +39,7 @@ describe('Auth Basics', function() {
             authorization: {
                 SUB: {
                     subscribe: "bar",
-                    publish: "bar"
+                    publish: ["bar"]
                 },
                 users: [{
                     user: 'bar',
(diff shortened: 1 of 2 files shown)


test/perms.js
```diff
@@ -87,4 +87,58 @@ describe('Auth Basics', function() {
...
+
+        nc.publish('bar');
+        nc.publish('foo');
+        nc.publish('bar');
+        nc.subscribe('foo', () => {});
+        nc.publish('bar');
+        nc.flush(() => {
+            nc.close();
+            if (errs.length !== 2) {
+                done(new Error("expected one permission error"));
```
andriy-bulynko: Should this be:
```suggestion
                done(new Error("expected two permission errors"));
```

aricart:
This matches the nats.ts handling (so nats.ts doesn't require a fix).

https://github.com/nats-io/nats.ts/blob/master/src/protocolhandler.ts#L642
https://github.com/nats-io/nats.ts/blob/master/src/protocolhandler.ts#L979

test/perms.js
aricart: Yes it should - thanks @andriy-bulynko

PR #293, opened 2019-09-04
FIX #292

### LINKED ISSUE #292: Argument handling
#286 contribution (thank you @basti1302)  pointed out that argument handling is currently too loose. That contribution addressed `publish`, but the same type of defect is apparent also on `subscribe`, `request`, etc.

aricart:
FIX #292 
Unlike nats.ts, nats.js has many optional arguments, and unfortunately the required arguments don't preceed optional ones, so argument shuffling is performed.

- This release will raise errors in dodgy code that was accepted by previous releases. For example a subscription or request that doesn't provide a callback will now fail, as a callback is required to process messages. Only optional callbacks can be omitted.

- In cases where a call has a callback, if a validation type error is raised, the error will be delivered through the callback provided. If the callback is not provided in optional cases, the error will be raised via the standard event emitter mechanism.

- This PR does away with deprecated functions.

### DIFF
--- a/test/shift.js
+++ b/test/shift.js
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2013-2019 The NATS Authors
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const NATS = require('../'),
+    should = require('should');
+
+describe('shifter', function() {
+    it('should handle empty', () => {
+        const a = new Array(0);
+        const b = NATS.callbackShifter(0, a);
+        b.should.have.length(0);
+    });
+
+    it('should handle just fun', () => {
+        const a  = [function () {}];
+        const b = NATS.callbackShifter(10, a);
+        b.should.have.length(10);
+        for(let i=0; i < 8; i++) {
+            (b[i] === undefined).should.be.true();
+        }
+        const v = typeof b[9];
+        v.should.be.equal('function');
+    });
+
+    it('should handle all', () => {
+        const a = ['a', '1', function () {}];
+        const b = NATS.callbackShifter(3, a);
+        should.deepEqual(a, b);
+    });
+
+    it('should handle holes', () => {
+        const a = ['a', '1', undefined, function () {}];
+        const b = NATS.callbackShifter(4, a);
+        should.deepEqual(a, b);
+    });
+});
--- a/test/subevents.js
+++ b/test/subevents.js
@@ -47,7 +47,7 @@ describe('Subscription Events', function() {
             nc.close();
             done();
         });
-        nc.subscribe(subj);
+        nc.subscribe(subj, () => {});
     });
 
     it('should generate subscribe events with opts', function(done) {
@@ -66,7 +66,7 @@ describe('Subscription Events', function() {
         });
         nc.subscribe(subj, {
             queue: queuegroup
-        });
+        }, () => {});
     });
 
     it('should generate unsubscribe events', function(done) {
@@ -79,7 +79,7 @@ describe('Subscription Events', function() {
             nc.close();
             done();
         });
-        var sid = nc.subscribe(subj);
+        var sid = nc.subscribe(subj, () => {});
         nc.unsubscribe(sid);
     });
 
@@ -95,7 +95,7 @@ describe('Subscription Events', function() {
         });
         nc.subscribe(subj, {
             max: 1
-        });
+        }, () => {});
         nc.publish(subj);
     });
 
@@ -105,10 +105,10 @@ describe('Subscription Events', function() {
         let eventsReceived = 0;
         const want = 5;
 
-        nc.on('unsubscribe', function(sid, subject) {
+        nc.on('unsubscribe', function() {
             eventsReceived++;
         });
-        const sid = nc.subscribe(subj);
+        const sid = nc.subscribe(subj, () => {});
         nc.unsubscribe(sid, want);
         for (let i = 0; i < want; i++) {
             nc.publish(subj);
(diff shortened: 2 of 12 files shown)


test/shift.js
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2013-2019 The NATS Authors
```
derekcollison: Looks like a new file.

Also does this have an impact on performance? Have we run those tests?

derekcollison: I mean the arg processing on all pub calls.

test/subevents.js
```diff
@@ -105,10 +105,10 @@ describe('Subscription Events', function() {
         let eventsReceived = 0;
         const want = 5;
 
-        nc.on('unsubscribe', function(sid, subject) {
+        nc.on('unsubscribe', function() {
             eventsReceived++;
         });
-        const sid = nc.subscribe(subj);
+        const sid = nc.subscribe(subj, () => {});
```
derekcollison: Seems sometimes you use function and other times the new syntax, Probably should be consistent.

basti1302:
Would be nice if this small change to the readme from my previous PR  could be salvaged, too:

https://github.com/nats-io/nats.js/pull/286/files#diff-04c6e90faac2675aa89e2176d2eec7d8R444-R446 

While this might be obvious to y'all, it might save people unfamiliar with the project a bit of head scratching.

test/subevents.js
aricart: need to change all the other uses before arrow was available to us. Yes it would simplify somethings, but also there are some differences.

test/shift.js
aricart: Yes indeed that has a performance.
```
PUBLISHER METRICS 
version                  max   min   samples  rate              
-----------------------  ----  ----  -------  ------------------
0.8.10 (string)          286   281   1000000  3,524,850 msgs/sec
1.3.0                    360   329   1000000  2,935,995 msgs/sec
2.0.0                    436   425   1000000  2,326,483 msgs/sec
0.8.10                   573   550   1000000  1,785,395 msgs/sec
1.0.0                    635   617   1000000  1,602,136 msgs/sec
1.0.0_nodelay            655   626   1000000  1,578,781 msgs/sec
0.8.10 - Buffer.alloc()  958   929   1000000  1,066,666 msgs/sec
0.8.10 - new Buffer()    1022  923   1000000  1,048,987 msgs/sec
0.8.6_1024_mcl           1060  999   1000000  986,635 msgs/sec  
0.8.6                    1026  1013  1000000  978,821 msgs/sec  

RECONNECT 
version                  max   min   samples  rate            
-----------------------  ----  ----  -------  ----------------
0.8.6_1024_mcl           2689  2191  1000000  407,479 msgs/sec
0.8.6                    2947  2168  1000000  406,305 msgs/sec
0.8.10                   3154  2321  1000000  380,966 msgs/sec
0.8.10 - Buffer.alloc()  2964  2337  1000000  378,028 msgs/sec
1.3.0                    3123  2611  1000000  362,318 msgs/sec
2.0.0                    3246  3018  1000000  319,437 msgs/sec

PUBLISH+SUBSCRIBER METRICS 
version                  max   min   samples  rate              
-----------------------  ----  ----  -------  ------------------
0.8.10                   930   881   1000000  1,112,594 msgs/sec
0.8.10 - Buffer.alloc()  929   887   1000000  1,104,972 msgs/sec
0.8.6                    916   900   1000000  1,100,715 msgs/sec
1.0.0_nodelay            939   928   1000000  1,073,422 msgs/sec
1.0.0                    944   926   1000000  1,072,271 msgs/sec
0.8.6_1024_mcl           1014  912   1000000  1,069,391 msgs/sec
1.3.0                    973   918   1000000  1,064,509 msgs/sec
2.0.0                    1074  1013  1000000  964,785 msgs/sec  

REQUEST REPLY METRICS 
version                  max  min  samples  lat                 rate            
-----------------------  ---  ---  -------  ------------------  ----------------
1.3.0                    547  542  100000   2.7191666666666667  183,879 msgs/sec
2.0.0                    625  583  100000   3.002222222222222   166,543 msgs/sec
0.8.10                   622  609  100000   3.0665000000000004  163,052 msgs/sec
0.8.10 - Buffer.alloc()  623  610  100000   3.075               162,601 msgs/sec
0.8.6                    759  725  100000                       136,314 msgs/sec
0.8.6_1024_mcl           760  741  100000                       133,511 msgs/sec
1.0.0                    765  745  100000   3.785               132,100 msgs/sec
1.0.0_nodelay            765  750  100000   3.7874999999999996  132,013 msgs/sec

SUBSCRIBER METRICS 
version                  max   min   samples  rate              
-----------------------  ----  ----  -------  ------------------
1.3.0                    687   639   1000000  1,523,229 msgs/sec
2.0.0                    1035  641   1000000  1,383,763 msgs/sec
0.8.10                   1036  887   1000000  1,032,844 msgs/sec
0.8.10 - Buffer.alloc()  1006  945   1000000  1,028,912 msgs/sec
0.8.6_1024_mcl           1071  1046  1000000  945,278 msgs/sec  
0.8.6                    1084  1059  1000000  935,692 msgs/sec 
```

aricart:
> Would be nice if this small change to the readme from my previous PR could be salvaged, too:

Indeed, I am going to put those changes in the current master, so that it makes it into the code base regardless of how long it takes for this PR to land.

aricart:
@derekcollison one concern with this PR is that it requires a major version bump. If that is the case, perhaps the correct fix, is to create an optional parameters object, and reshuffle the order of the required params. That do away with parameter shifting and inspection. The param will be set or not. I personally think that would be a more forward looking change.

test/shift.js
derekcollison: What should I be looking at specifically to compare this exact change?

test/subevents.js
derekcollison: Do you agree we should try to be consistent and use one or another and not mix and match?

derekcollison:
Would it break the majority of current programs?

aricart:
> Would it break the majority of current programs?

Doggy ones for sure. Like our tests, we set up fake subscriptions without a handler, etc.
Also notifications will always be delivered via the callback if provided.

derekcollison:
I meant about your last comment about re-ordering required arguments.

aricart:
> I meant about your last comment about re-ordering required arguments.

For example 

publish(subject, reply, function) => publish(subject, data, opt_options)
// opt_options =  {reply:, callback:}
// data would be required, but it can always be "", null, undefined

subscribe(subject, opts, callback) => subscribe(subject, callback, opt_options)
// where opts include timeout, max, queue, etc)
// we have the opportunity to have subscription be an actual object rather than dealing with ids.
// eliminate timeout(), since that could be a subscription option

request(subject, data, opts, callback) => request(subject, data, callback, opt_options)
// could get rid of requestOne - since request could just get one unless a max number of messages is specified

derekcollison:
Looking over your suggestion I think it's correct for a 2.0 version. Appreciate the effort.

aricart:
closing as 2.0 addresses all the issues discussed here.

PR #297, opened 2019-09-24
Transposition

aricart:
This is a similar error to the one we were discussing on #292. Effectively, this is another upvote.
The fix is c/p from #292. Before the end of the year we need to cut a 2.0, with the changes discussed.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1884,7 +1884,6 @@ Client.prototype.requestOne = function(subject, opt_msg, opt_options, timeout, c
     if (this.options.useOldRequestStyle) {
         return this.oldRequestOne(subject, opt_msg, opt_options, timeout, callback);
     }
-
     // eslint-disable-next-line prefer-rest-params
     const args = callbackShifter(5, Array.prototype.slice.call(arguments));
     subject = args[0];
@@ -1897,6 +1896,7 @@ Client.prototype.requestOne = function(subject, opt_msg, opt_options, timeout, c
         timeout = opt_options;
         opt_options = undefined;
     }
+
     if (timeout === undefined && opt_options === undefined && typeof opt_msg === "number") {
         timeout = opt_msg;
         opt_msg = EMPTY;
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1115,8 +1115,9 @@ Client.prototype.processInbound = function() {
                 } else if ((m = OK.exec(buf)) !== null) {
                     // Ignore for now..
                 } else if ((m = ERR.exec(buf)) !== null) {
-                    this.processErr(m[1]);
-                    return;
+                    if (this.processErr(m[1])) {
+                        return;
+                    }
                 } else if ((m = PONG.exec(buf)) !== null) {
                     this.pout = 0;
                     const cb = this.pongs && this.pongs.shift();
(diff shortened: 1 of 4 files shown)


test/basics.js
```diff
@@ -877,4 +877,87 @@ describe('Basics', function() {
...
+
+            nc.flush(() => {
+                nc.requestOne(subj, 'some string', {queue: 'worker'}, 1000, h);
+                nc.requestOne(subj, {}, {queue: 'worker'}, 1000, h);
+                nc.requestOne(subj, 123, {queue: 'worker'}, 1000, h);
+            });
+        });
+    });
+
+    it('json requests', function(done) {
```
kozlovic: Looks to me that this is the exact same test than the one just above.

test/shifter.js
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 019 The NATS Authors
```
kozlovic: Date is wrong

test/shifter.js
```diff
@@ -0,0 +1,50 @@
...
+        const a = new Array(0);
+        const b = NATS.callbackShifter(0, a);
+        b.should.have.length(0);
+    });
+
+    it('should handle just fun', () => {
+        const a  = [function () {}];
+        const b = NATS.callbackShifter(10, a);
+        b.should.have.length(10);
+        for(let i=0; i < 8; i++) {
```
kozlovic: `<=8`

lib/nats.js
```diff
@@ -1905,6 +1909,40 @@ Client.prototype.requestOne = function(subject, opt_msg, opt_options, timeout, c
     return this.request(subject, opt_msg, opt_options, callback);
 };
 
+// finds the last index that satisfies the testFun in array a
+function lastIndexOf(a, testFun) {
+    const idx = a.length-1;
+    for(let i=idx; i > -1; i--) {
```
kozlovic: `i >= 0`?

lib/nats.js
```diff
@@ -1905,6 +1909,40 @@ Client.prototype.requestOne = function(subject, opt_msg, opt_options, timeout, c
...
+// returns an array that shifts the last function in the
+// array to the last element of the array
+function callbackShifter(maxArgs, args) {
+    // last argument is expected to be a function
+    const a = new Array(maxArgs);
+    const idx = lastIndexOf(args, (v) => typeof v === 'function');
+
+    // found a function, at the last index
+    if (idx !== -1) {
+        a[maxArgs - 1] = args[idx];
```
kozlovic: Looks to me that you would lose the last element if that is not a function, and you would still keep the function at its original position. Suppose you have:
`1, 2, 3, F, 4, 5` where `F` is a function. The `lastIndexOf()` would return idx 3. In array `a` you would put `a[5]='F'`, and then copy only from 0 to 2, so `a` would result in: `1, 2, 3, _,  _, F` if I understand your code.. (here the `_` means that there is no value copied from `args`, so undefined I guess.

aricart: the callbacks always end in a function, so arguments 4, 5 and not valid, and are dropped - this is is correct.

kozlovic: but maxArgs is 6 in the above case, no?

PR #298, opened 2019-09-24
FIX #292 - json requestOne fix

### LINKED ISSUE #292: Argument handling

aricart:
For JSON clients, a number parameter could be a value. The NATS library incorrectly took the first number as the timeout.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -1884,18 +1884,22 @@ Client.prototype.requestOne = function(subject, opt_msg, opt_options, timeout, c
     if (this.options.useOldRequestStyle) {
         return this.oldRequestOne(subject, opt_msg, opt_options, timeout, callback);
     }
+    // eslint-disable-next-line prefer-rest-params
+    const args = callbackShifter(5, Array.prototype.slice.call(arguments));
+    subject = args[0];
+    opt_msg = args[1];
+    opt_options = args[2];
+    timeout = args[3];
+    callback = args[4];
 
-    if (typeof opt_msg === 'number') {
-        callback = opt_options;
-        timeout = opt_msg;
-        opt_options = null;
-        opt_msg = EMPTY;
+    if (timeout === undefined && typeof opt_options === "number") {
+        timeout = opt_options;
+        opt_options = undefined;
     }
 
-    if (typeof opt_options === 'number') {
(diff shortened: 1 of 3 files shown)


aricart:
@kozlovic  sorry - I had a git explosion trying to get it to rebase, so I made a new PR.
I have updated the tests to cover the possible 'correct' number of arguments

lib/nats.js
```diff
@@ -1884,27 +1884,65 @@ Client.prototype.requestOne = function(subject, opt_msg, opt_options, timeout, c
...
+    const idx = lastIndexOf(args, (v) => typeof v === 'function');
+
+    // found a function, at the last index
+    if (idx !== -1) {
+        a[maxArgs - 1] = args[idx];
+        for (let i = 0; i < idx; i++) {
+            a[i] = args[i];
+        }
+    } else {
+        for (let i=0; i < args.length; i++) {
```
kozlovic: Did you not say that you could return args here?

test/basics.js
```diff
@@ -877,4 +877,117 @@ describe('Basics', function() {
...
+
+                nc.requestOne(subj, 'a', {queue: 'worker'}, 1000, h);
+                nc.requestOne(subj, {}, {queue: 'worker'}, 1000, h);
+                nc.requestOne(subj, 10, {queue: 'worker'}, 1000, h);
+            });
+        });
+    });
+
+    it('json json old requests', function(done) {
+        const nc = NATS.connect({port: PORT, json: true, useOldRequestStyle: true});
```
kozlovic: Would have been nice to have a function that is shared by those 2 `if()` tests and simply ass the `useOldRequestStyle` option to that common function.

lib/nats.js
aricart: rethought that at least we get a sanitized version of the args - (it will be the expected length regardless of what they passed)

test/basics.js
aricart: I'll refactor all that soon - and get the deprecated junk out of there.

PR #306, opened 2020-01-07
standard-js

aricart:
adoption of standard-js for the formatting.
this change removes the use of url.parse() (bumping to v2.0.0)
bumped min requirements for node to 10, matching currently supported versions


benchmark/pub_perf.js
```diff
@@ -13,52 +13,49 @@
...
 
-const NATS = require('../lib/nats');
-const nats = NATS.connect();
-const fs = require('fs');
+const NATS = require('../lib/nats')
+const nats = NATS.connect()
+const fs = require('fs')
 
-///////////////////////////////////////
+/// ////////////////////////////////////
```
derekcollison: This looks weird, why did it add that?

benchmark/pub_sub_perf.js
```diff
@@ -13,56 +13,54 @@
...
-const NATS = require('../lib/nats');
-const nc1 = NATS.connect();
-const nc2 = NATS.connect();
+const fs = require('fs')
+const NATS = require('../lib/nats')
+const nc1 = NATS.connect()
+const nc2 = NATS.connect()
 
-///////////////////////////////////////
+/// ////////////////////////////////////
```
derekcollison: Same here.

benchmark/reconnect_perf.js
```diff
@@ -13,62 +13,61 @@
...
-const fs = require('fs');
-const NATS = require('../lib/nats'),
-    nsc = require('../test/support/nats_server_control');
+const fs = require('fs')
+const NATS = require('../lib/nats')
+const nsc = require('../test/support/nats_server_control')
 
-
-///////////////////////////////////////
+/// ////////////////////////////////////
```
derekcollison: Confused why it keeps doing this.

benchmark/pub_perf.js
aricart: @derekcollison 'spaced-comment' rule:
https://eslint.org/docs/rules/spaced-comment

The standardjs is opinionated like go, and they don't like that :(

derekcollison: ok

aricart:
@derekcollison lowered version to 1.3.6, as `new URL()` is available since 7.0.0, I am going to do a test, I might have to add an import instead of using a global object.

benchmark/pub_perf.js
kozlovic: Yes, but you could still replace with another valid comments maybe?
Like:
```
/*
 * Publish Performance
 *
 */
```
or some others in the list of "correct" comments?

kozlovic: Or this maybe?
```
//--------------
// Publisher Performance
//---------------
```

aricart:
Yes - changed so that we `require('url')` and do `new url.URL()` as this is compatible all the way back to node 7.

PR #307, opened 2020-01-10
Readme, Example, and typescript binding cleanup

aricart:
- fixed examples usage of `--creds` parameter
- added support for `--queue` group to node-sub
- added tool `node-rep` for botting `node-req` uses
- made the `node-[pub|sub|req|rep]` example tools binaries, this allows developers to use these tools while developing their apps with `npx node-[pub|sub|req|rep]` without additional copying or editing

- unified source code version with its package.json version - this allows the npm version management to work without breaking the tests or requiring manual edit

- simplified typescript definitions for `connect`, definition was missing hybrid `connect(url, opts)`.
- added typescript type `RequestOptions` for request options, these were incorrectly referencing `SubscribeOptions`

- reworked README.md examples formatting, removed example of `requestOne` as it provides no benefit of over `request` when additional arguments are specified. A future version should deprecate this API.


README.md
```diff
@@ -227,166 +215,171 @@ const sid1 = nc.subscribe(subj, {queue: 'q1'}, () => {
...
 
 ```javascript
-var nats = require('nats');
-var fs = require('fs');
+const NATS = require('nats')
+const fs = require('fs')
 
 // Simple TLS connect
-var nc = nats.connect({port: TLSPORT, tls: true});
+let nc = NATS.connect({ tls: true })
```
derekcollison: Do we need this? If server requires TLS it should just do the right thing yes?

examples/node-pub
```diff
@@ -18,7 +18,8 @@
 /* jslint node: true */
 'use strict'
 
-const nats = require('nats')
+// change '../' to 'nats' when copying
```
derekcollison: Not sure what this means.

examples/node-pub
```diff
@@ -27,19 +28,19 @@ const subject = argv._[0]
 const msg = argv._[1] || ''
 
 if (!subject) {
-  console.log('Usage: node-pub  [-s server] [--creds file] <subject> [msg]')
+  console.log('Usage: node-pub  [-s server] [--creds=filepath] <subject> [msg]')
```
derekcollison: X=Y syntax required?

examples/node-rep
```diff
@@ -0,0 +1,69 @@
...
+const argv = require('minimist')(process.argv.slice(2))
+
+const url = argv.s || nats.DEFAULT_URI
+const creds = argv.creds
+const queue = argv.queue
+const subject = argv._[0]
+const response = argv._[1] || ''
+
+if (!subject) {
+  console.log('Usage: node-rep  [-s server] [--creds=creds_file_path] [--queue=qname] <subject> [response]')
```
derekcollison: Made mw think of repl, had to read code. Would make node-reply or nats-reply.

README.md
aricart: this is saying that the client requires the connection to be tls.

examples/node-pub
aricart: I made the examples work right from the package when installed as `npx

examples/node-pub
aricart: yes for flags that are not a single letter

examples/node-rep
aricart: Let me rename it

derekcollison: ok

PR #311, opened 2020-01-19
NATS#connect - prevent tls.connect from throwing uncaught exceptions (fixes #310)

### LINKED ISSUE #310: NATS#connect - missing error handling for offline TLS options' verification [Bug]
## Introduction
When TLS options are specified under the `tls` key, the NATS client attempts to open a [`tlsSocket`](https://nodejs.org/api/tls.html#tls_class_tls_tlssocket).

If the provided options are invalid as is the case in the following examples, there is no error handling which warrants Node to emit a `unhandledException` event which can be dangerous for the consumers.

1) if the `key` does not match the `cert`;
2) if the `key` or the `cert` is malformed; etc.

These errors are thrown as a part of the sanity check in OpenSSL and take place **before** a connection has been initialized.

## Bug
If we had a stream, its event listeners are stripped
https://github.com/nats-io/nats.js/blob/407193dbe851ac336da4c9cff1e052ffd9328bdf/lib/nats.js#L1170

TLS options are sanitized and sent over to `tls.connect()` here:
https://github.com/nats-io/nats.js/blob/407193dbe851ac336da4c9cff1e052ffd9328bdf/lib/nats.js#L1172

After this, event handlers are re-attached with:
https://github.com/nats-io/nats.js/blob/407193dbe851ac336da4c9cff1e052ffd9328bdf/lib/nats.js#L1175

The error handling works perfectly for errors happening after the initial handshake (SNI, etc.) However, OpenSSL does a sanity check on the CA, Certificates and their keys amongst other things **before** starting a TCP connection. These errors are `throw`n as opposed to emitted in a wrapped event. Since we do not have **any** error handling for thrown events, this leads to a `unhandledException` scenario.

## Reproduction Steps
The following options configurations will trigger the bug:

### Case 1
```javascript
const fs = require( 'fs' );
const nats = require( 'nats' );

const options = {
    url: 'nats://127.0.0.1:4222',
    tls: {
        key: fs.readFileSync( path.join( __dirname, './client-1.key' ) ),
        cert: fs.readFileSync( path.join( __dirname, './client-2.crt' ) ), 
        ca: [ fs.readFileSync( path.join( __dirname, './ca.crt' ) ) ],
    }
};

const _client = nats.connect( options );
_client.on( 'error', error => {
    console.trace( 'There was an error: %s', error.message );
} );

_client.on( 'connect', () => {
    _client.close();
    console.log( 'Client connected!' );
} );
```

### Case 2
```javascript
const fs = require( 'fs' );
const nats = require( 'nats' );

const options = {
    url: 'nats://127.0.0.1:4222',
    tls: {
        key: fs.readFileSync( path.join( __dirname, './client-2.key' ) ),
        cert: fs.readFileSync( path.join( __dirname, './client-1.crt' ) ), 
        ca: [ fs.readFileSync( path.join( __dirname, './ca.crt' ) ) ],
    }
};

const _client = nats.connect( options );
_client.on( 'error', error => {
    console.trace( 'There was an error: %s', error.message );
} );

_client.on( 'connect', () => {
    _client.close();
    console.log( 'Client connected!' );
} );
```

These cases should throw a `ERR_OSSL_X509_KEY_VALUES_MISMATCH` error.

## Mitigation
The mitigation is _very_ straightforward with just a simple `try-catch` block around `tls.connect(...)`:

```javascript
try {
    this.stream = tls.connect(tlsOpts, () => {
        this.flushPending();
    });
} catch ( error ) {
    this.emit('error', new NatsError(CONN_ERR_MSG_PREFIX + error, CONN_ERR, error));
}
```

This seems to take care of the error and gives a well-formatted error to the consumer.

## Conclusion
The test cases for this are pretty simple as is the implementation; as such, if it's agreed to by the community, I'd be more than happy to submit a PR for the issue at hand. 💯 😄

labsvisual:
## Introduction
As described in #310, `tls.connect` throws an uncaught exception if OpenSSL's pre-flight verification fails. This pull request fixes that.

## Description of the Fix
A simple `try-catch` wrapper around `tls.connect` in `lib/nats.js` seems to do the trick just fine. For additional semantic interoperability, a new error code definition (and its accompanying message prefix) was added; namely:
* `OPENSSL_ERR` - set to `OPENSSL_ERR`
* `OPENSSL_ERR_MSG_PREFIX` - set to `TLS credentials verification failed: `

## Breaking Changes
None.

## Compatibility Issues
None.

## Testbeds
* `Darwin` - `18.7.0` 
* [Travis](https://travis-ci.com/labsvisual/nats.js/builds/145139786)

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -100,6 +100,8 @@ const SUB_DRAINING = 'SUB_DRAINING'
 const SUB_DRAINING_MSG = 'Subscription draining'
 const CONN_ERR = 'CONN_ERR'
 const CONN_ERR_MSG_PREFIX = 'Could not connect to server: '
+const OPENSSL_ERR = 'OPENSSL_ERR'
+const OPENSSL_ERR_MSG_PREFIX = 'TLS credentials verification failed: '
 const INVALID_ENCODING = 'INVALID_ENCODING'
 const INVALID_ENCODING_MSG_PREFIX = 'Invalid Encoding:'
 const NATS_PROTOCOL_ERR = 'NATS_PROTOCOL_ERR'
@@ -1169,9 +1171,15 @@ Client.prototype.processInbound = function () {
               if (this.stream) {
                 this.stream.removeAllListeners()
               }
-              this.stream = tls.connect(tlsOpts, () => {
-                this.flushPending()
-              })
+              try {
+                // Refer to issue #310
+                this.stream = tls.connect(tlsOpts, () => {
+                  this.flushPending()
+                })
+              } catch (error) {
+                this.emit('error', new NatsError(OPENSSL_ERR_MSG_PREFIX + error, OPENSSL_ERR, error))
+                return
+              }
               this.setupHandlers()
             }
 
--- a/test/tls.js
+++ b/test/tls.js
@@ -143,4 +143,44 @@ describe('TLS', function () {
       done()
     })
   })
+
+  describe('OpenSSL preflight verification errors', function () {
+    const testTable = [
+      {
+        errorCode: 'ERR_OSSL_X509_KEY_VALUES_MISMATCH',
+        tls: {
+          key: fs.readFileSync('./test/certs/client-key.pem'),
+          cert: fs.readFileSync('./test/certs/ca.pem'),
+          ca: [fs.readFileSync('./test/certs/server.pem')]
+        }
+      },
+      {
+        errorCode: 'ERR_OSSL_PEM_NO_START_LINE',
+        tls: {
+          key: fs.readFileSync('./test/certs/client-cert.pem'),
+          cert: fs.readFileSync('./test/certs/client-key.pem'),
+          ca: [fs.readFileSync('./test/certs/ca.pem')]
+        }
+      }
+    ]
+
+    testTable.forEach(({ errorCode, tls }) => {
+      it(`should handle ${errorCode}`, function (done) {
+        const nc = NATS.connect({
+          port: TLSPORT,
+          tls
+        })
+        nc.once('connect', function (done) {
+          done(new Error(`was expecting a ${errorCode} OpenSSL error`))
+        })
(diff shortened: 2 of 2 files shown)


aricart:
@labsvisual  I wish all PRs came like this - awesome contribution thanks.

labsvisual:
@aricart - all thanks to you; super simple to understand the codebase and extremely quick to get up and running with it!

If possible, I'd love to contribute more to this project (either in Node.js or Golang) 😃 

---

On a completely different note, I did find another issue (purely related to bindings and not the functionality.) If we look at the TypeScript bindings in the file `index.d.ts` at line 33:
https://github.com/nats-io/nats.js/blob/d9270049f7948bbe3d67db37b0717172cfe4d33c/index.d.ts#L33

line 37
https://github.com/nats-io/nats.js/blob/d9270049f7948bbe3d67db37b0717172cfe4d33c/index.d.ts#L37

and line 40
https://github.com/nats-io/nats.js/blob/d9270049f7948bbe3d67db37b0717172cfe4d33c/index.d.ts#L40

We see some exports defined by the bindings which are not found in the actual codebase; I did a quick search and we seem to only export `NATS_PROTOCOL_ERR`:
https://github.com/nats-io/nats.js/blob/d9270049f7948bbe3d67db37b0717172cfe4d33c/lib/nats.js#L174

This might get a little confusing for a developer. Would you mind a quick PR from me to resolve this or is this something which has been explicitly included?

derekcollison:
We would love more contributions!

labsvisual:
On it! @aricart, @derekcollison - anything specific you would want me to start with?

derekcollison:
If there is something you already have in mind, run with that. Otherwise I would take a look at the issues list for each repo for some ideas.

aricart:
@labsvisual  simply make a new PR for those constants! Thanks.

PR #315, opened 2020-01-22
Connect timeout

### LINKED ISSUE #309: NATS#connect - specify socket timeout options [Feature Request]
## Introduction
The `NATS.connect()` method, when given an unreachable IP address, times out after the default 120s (2 minutes). However, this time period is too long for constantly changing microservices and their respective upstream NATS transporters. There should be a way for a client to specify the timeout manually as a property to the options passed to the `.connect()` method.

## Rationale
In an environment where connection parameters are provided through a reactive KV store (etcd, Consul, etc.) the application should have the ability to know when a given Host/IP and port pair is unreachable in a specific amount of time lest incurring a downtime of no visibility for at least 2 minutes.

Consider a case where the configuration parameters have just been changed and the services are reloading; for 2 minutes, they will just hang with no output thus not giving any real telemetry data pointing to am erred configuration option.

## Implementation
In order to understand the implementation, we must first look at the line which actually initializes a TCP connection: 

https://github.com/nats-io/nats.js/blob/407193dbe851ac336da4c9cff1e052ffd9328bdf/lib/nats.js#L869

`net.createConnection(port, host)` calls `socket.connect(port, host)` wherein the default socket options (inter alia, `timeout` set to `120000`) are used.
### Caveat
The `net` module does not provide a method to set the connect timeout; the `socket.setTimeout` sets the time before which the socket is destroyed when no data is received or sent (basically it sets the `SO_RCVTIMEO` and `SO_SNDTIMEO` parameters).

### Basic Implementation
We can abuse how `.setTimeout()` works and wrap the connect method as a deferred function call. 

Something like this, probably:

```javascript
const net = require( 'net' );

const socket = net.createConnection( port, host );
const _timeout = setTimeout( () => {
    socket.destroy();
    throw new errors.SocketTimeoutError( options.timeout );
}, options.timeout );

socket.on( 'connect', () => clearTimeout( _timeout ) );
this.stream = socket;
```

The idea here is that if we do not receive the `connect` event after `options.timeout` milliseconds, the function in the `setTimeout()` call will destroy the socket and throw an error which the client can use to understand what happened. If, however, the socket manages to connect within the specified period, we just clear the timeout thus preventing it from throwing any errors. 

Please keep in mind that this is just a very basic PoC-like approach and might require some test cases to ensure cross-platform compatibility.

## Conclusion
That is what I had in mind. I'd be more than happy to submit a PR for this should this feature be accepted. 😄

aricart:
FIX #309 - Added `timeout` connection option. The `timeout` option specifies the number of millis to wait for a connection to occur and for data to be sent by the server. The client must complete the dial and receive a `PONG` from the server within this amount of time, or it fails.

Fixed a small issue where if an options argument was provided but was not an object, the client would
try to emit an error, however, the emit mechanism was not yet set up, as the client was not yet created.

Also fixed an issue where if the client emitted an error (if it failed to connect to a server, the client would close the stream, but the client was not closed and thus preventing the node application from terminating.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -775,7 +775,8 @@ Client.prototype.setupHandlers = function () {
   })
 
   stream.on('data', (data) => {
-    if (this.options.timeout) {
+    if (toh) {
+      toh = undefined
       // connect added a timeout on data from the socket - remove it
       stream.setTimeout(0)
     }
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -707,22 +707,13 @@ Client.prototype.setupHandlers = function () {
     return
   }
 
-  let toh
   if (this.options.timeout) {
-    toh = setTimeout(() => {
+    this.connectionTimeoutHandler = setTimeout(() => {
       stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
     }, this.options.timeout)
   }
 
   stream.on('connect', () => {
-    if (toh) {
-      // we got a connection
-      clearTimeout(toh)
-      // now we hope to get data on the socket before the timeout
-      stream.setTimeout(this.options.timeout, () => {
-        stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
-      })
-    }
     this.cancelHeartbeat()
     this.connected = true
     this.scheduleHeartbeat()
@@ -775,11 +766,6 @@ Client.prototype.setupHandlers = function () {
   })
 
   stream.on('data', (data) => {
-    if (toh) {
-      toh = undefined
-      // connect added a timeout on data from the socket - remove it
-      stream.setTimeout(0)
-    }
     // If inbound exists, concat them together. We try to avoid this for split
     // messages, so this should only really happen for a split control line.
     // Long term answer is hand rolled parser and not regexp.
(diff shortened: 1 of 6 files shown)


lib/nats.js
```diff
@@ -742,14 +764,21 @@ Client.prototype.setupHandlers = function () {
...
+      return
     }
+    // continue with reconnect
     this.closeStream()
   })
 
   stream.on('data', (data) => {
+    if (this.options.timeout) {
+      // connect added a timeout on data from the socket - remove it
+      stream.setTimeout(0)
```
kozlovic: This is going to call this for every incoming message, does not seem efficient.

lib/nats.js
```diff
@@ -700,7 +707,22 @@ Client.prototype.setupHandlers = function () {
     return
   }
 
+  let toh
```
kozlovic: Why would you set a regular timeout since you have access to `stream`, why not use `stream.setTimeout()` here instead of in `stream.on('connect')`? Would setting the streaming.setTimeout() before "on connect" would not cover the connect itself taking too much time?
If you have to set this non socket-level timer, then I would set it once here and clear it only when the PONG to initial PING has been received.

test/timeouts.js
```diff
@@ -36,6 +37,73 @@ describe('Timeout and max received events for subscriptions', function () {
...
+        srv.close(done)
+      })
+      nc.on('error', (err) => {
+        done(new Error('shouldnt have failed: ' + err))
+      })
+    })
+  }, 15000)
+
+  it('connection timeout doesnt trigger', (done) => {
+    // this is not expecting to have enough time to connect
```
kozlovic: I think this comment is wrong? You are expecting the connect to have enough time and check that no connection timeout error is raised, right?

lib/nats.js
aricart: there's no connect timeout, the other timeout there's no handle for it

lib/nats.js
aricart: zonk!

test/timeouts.js
aricart: I am expecting that in 1ms, it is not going to connect to the remote system, so it triggers the first condition.

kozlovic: No, this test passes 1000 (1sec) and I think expect to be able to connect and make sure that this is not raising the timeout error..

lib/nats.js
aricart: Also, having a timer removed when it is processing the pings, etc, could trigger an unknown amount of time, I believe the correct thing to do is to tighten the check to where it matters.

test/timeouts.js
aricart: c/p it is the lower function.

PR #324, opened 2020-02-06
FIX #323 - Remove Noop code collecting pending PUB and Pings

### LINKED ISSUE #323: remove scanning of PUB message on reconnect, as the pending buffer is dropped on disconnect.

aricart:
Removed noop code that collected pending PUB and PINGs. The code was a noop, because on any disconnect, `closeStream()` would already reset the pending buffer and the pongs. So nothing was recovered. 

As before until there's a connect, any contents in the pending is preserved and sent should it connect.
This change is no different than the current behavior, except that the previous code eluded to something that never happened.

This behavior lines up with go and other clients.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -89,6 +89,7 @@ const CONN_CLOSED = 'CONN_CLOSED'
 const CONN_DRAINING = 'CONN_DRAINING'
 const CONN_ERR = 'CONN_ERR'
 const CONN_TIMEOUT = 'CONN_TIMEOUT'
+const DISCONNECT_ERR = 'DISCONNECT'
 const INVALID_ENCODING = 'INVALID_ENCODING'
 const NATS_PROTOCOL_ERR = 'NATS_PROTOCOL_ERR'
 const NKEY_OR_JWT_REQ = 'NKEY_OR_JWT_REQ'
@@ -120,6 +121,7 @@ const CONN_CLOSED_MSG = 'Connection closed'
 const CONN_DRAINING_MSG = 'Connection draining'
 const CONN_ERR_MSG_PREFIX = 'Could not connect to server: '
 const CONN_TIMEOUT_MSG = 'Connection timeout'
+const DISCONNECT_MSG = 'Client disconnected, flush was reset'
 const INVALID_ENCODING_MSG_PREFIX = 'Invalid Encoding:'
 const NKEY_OR_JWT_REQ_MSG = 'An Nkey or User JWT callback needs to be defined.'
 const NON_SECURE_CONN_REQ_MSG = 'Server does not support a secure connection.'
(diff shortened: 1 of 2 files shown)


lib/nats.js
```diff
@@ -846,40 +846,11 @@ Client.prototype.sendConnect = function () {
...
-        pend.push(cmd)
-        pSize += cmdLen
-      }
-    })
-  }
-  this.pongs = pong
-  this.pending = pend
-  this.pSize = pSize
+  // a disconnect will prune any pending state - if we have data
+  // is because client has added messages to pending after the disconnect
```
derekcollison: it's

aricart:
> Are we sure this code was not doing something needed?

When the socket closes, the handler:
```
  stream.on('close', () => {
    this.closeStream()
...
``` 
Which then resolved to 
```
Client.prototype.closeStream = function () {
  if (this.stream !== null) {
    this.stream.destroy()
    this.stream = null
  }
  if (this.connected === true || this.closed === true) {
    this.pongs = null
    this.pout = 0
    this.pending = null
    this.pSize = 0
    this.connected = false
  }
  this.inbound = null
}
```
So it copied, but none of the contents of the pending buffer ever actually made it out.

Mikxail:
`closeStream` resolver are incorrect. Publish callbacks will be lost.
Test:
```
it('should properly do a publish callback if published after disconnect', (done) => {
    const nc = NATS.connect({
      port: PORT,
      reconnectTimeWait: 100
    })
    nc.subscribe('foo')
    nc.on('disconnect', (client) => {
      // restart server
      if (server === null) {
        nc.publish('foo', (err) => {
//          should.not.exist(err)
          nc.close()
          done()
        })
        server = nsc.startServer(PORT)
      }
    })
    nc.on('connect', () => {
      const s = server
      server = null
      nsc.stopServer(s)
    })
  })
```

aricart:
reset the branch as we had a lot of changes in the middle.
This gets rid of the no-op code as before, but we fail any pending flush callbacks that were present after the disconnect. This at least gives insight that something was not quite right. Any flushes, pubs after the disconnect are preserved until the next reconnect.

aricart:
> `closeStream` resolver are incorrect. Publish callbacks will be lost.
> Test:
> 
> ```
> it('should properly do a publish callback if published after disconnect', (done) => {
>     const nc = NATS.connect({
>       port: PORT,
>       reconnectTimeWait: 100
>     })
>     nc.subscribe('foo')
>     nc.on('disconnect', (client) => {
>       // restart server
>       if (server === null) {
>         nc.publish('foo', (err) => {
>           nc.close()
>           done()
>         })
>         server = nsc.startServer(PORT)
>       }
>     })
>     nc.on('connect', () => {
>       const s = server
>       server = null
>       nsc.stopServer(s)
>     })
>   })
> ```

Yes added a test at the protocol level

Mikxail:
@aricart That PR fixed most of reconnecting issues. But some of issue still exists.
`flushPending` method flushing `pending`s buffers right after try to write to the stream. But writing operation can failed. As I can suggest, nats client should not flush `penging`s messages until got response from nats server(except "noEcho" server mode).

You can simple verify issue, just modify your test:
```js
      const p = srv.address().port
      const nc = NATS.connect('nats://localhost:' + p, {
        reconnect: true,
        reconnectTimeWait: 250
      })
      // added stub for "write" method
      const stub1 = sinon.stub(nc.stream, 'write').callsFake(data => {
        if (/^PING/.test(data)) {
          socket.destroy()
          stub1.restore()
        }
      }).callThrough()
      nc.on('connect', () => {
        nc.pongs.push((err) => {
          if (err) {
            flushErr = true
          }
        })
        // process.nextTick(() => {
        //   // this is going to fake an outstanding ping
        //   socket.destroy()
        // })
      })
```

PR #328, opened 2020-02-11
Normalize behaviour of `reconnectTimeWait` and `maxReconnects` to match other clients.

### LINKED ISSUE #318: behavior of maxReconnectAttempts does not match go client
The go client counts reconnect attempts per server it knows about. Say the client knows about 3 servers and max reconnect attempts is 2. The client will attempt to connect each server twice and perform a total of 6 connect attempts.

This is in contrast to, the js client and presumably java client. They behave like this (excerpt from doc, will change)
Applications can set the maximum reconnect attempts. Generally, this will limit the actual number of attempts total, but check your library documentation. For example, in Java, if the client knows about 3 servers and the maximum reconnects is set to 2, it will not try all of the servers. On the other hand, if the maximum is set to 6 it will try all of the servers twice before considering the reconnect a failure and closing.

This seems to be an issue across clients...

### LINKED ISSUE #319: sleeps for reconnectTimeWait on every reconnect. this differs from other clients
Other clients look at time since last connect attempt to decide wether to delay or not.

https://github.com/nats-io/nats.js/blob/54d7e730eeb9be0f2701433a33cf2e1810179558/lib/nats.js#L2188-L2195

aricart:
FIX #319 sleeps for reconnectTimeWait on every reconnect this differs from other clients
FIX #318 reconnect behaviour only tries maxReconnects, rather than maxReconnects per server
Removed unreferenced variable 'server' from client
Removed unreferenced callback variables

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -210,6 +210,8 @@ function Client (opts) {
   events.EventEmitter.call(this)
   this.parseOptions(opts)
   this.initState()
+  // Select a server to connect to.
+  this.selectServer()
   this.createConnection()
 }
 
@@ -437,6 +439,7 @@ function Server (url) {
   this.url = url
   this.didConnect = false
   this.reconnects = 0
+  this.lastConnect = 0
 }
 
 /**
@@ -484,6 +487,7 @@ Client.prototype.selectServer = function () {
     }
   }
   this.servers.push(server)
+  return server
 }
 
 Client.prototype.checkNoEchoMismatch = function () {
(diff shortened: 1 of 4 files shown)


lib/nats.js
```diff
@@ -2200,7 +2202,46 @@ Client.prototype.scheduleReconnect = function () {
   if (this.servers[0].didConnect === true) {
     wait = this.options.reconnectTimeWait
   }
+  // Select a server to connect to - this will be
+  // the first server that meets the reconnectTimeWait criteria
+  // pick a ceiling number for the max we will wait
+  const now = Date.now()
+  let maxWait = wait
+  // some big number
```
kozlovic: What is this comment?

aricart: a left over from a different strategy

PR #338, opened 2020-03-25
Fix connect timeout occuring more than once

CaptainYarb:
Fix connect timeout not destroying correct reference to net stream


CaptainYarb:
oops, forgot to create a PR branch.

PR #340, opened 2020-03-25
Double timeout and socket.destroy() not issuing close.

aricart:
- Checked for a possible timer already scheduled (via the first connection, prior to tls upgrade) and removed the timeout in that case.

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -713,14 +713,13 @@ Client.prototype.setupHandlers = function () {
     return
   }
 
-  if (this.connectionTimeoutHandler) {
-    clearTimeout(this.connectionTimeoutHandler)
-    delete this.connectionTimeoutHandler
-  }
-
-  if (this.options.timeout) {
+  // tls upgrade will re-bind the handlers here, so if we have a timer we are going to ignore it
+  if (this.options.timeout && !this.connectionTimeoutHandler) {
     this.connectionTimeoutHandler = setTimeout(() => {
-      stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
+      if (this.stream) {
+        // this fires on the current stream
+        this.stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
+      }
     }, this.options.timeout)
   }
 
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -713,6 +713,11 @@ Client.prototype.setupHandlers = function () {
     return
   }
 
+  if (this.connectionTimeoutHandler) {
+    clearTimeout(this.connectionTimeoutHandler)
+    delete this.connectionTimeoutHandler
+  }
+
   if (this.options.timeout) {
     this.connectionTimeoutHandler = setTimeout(() => {
       stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -702,6 +702,16 @@ Client.prototype.scheduleHeartbeat = function () {
 }
 
 /**
+ * @api private
+ */
+Client.prototype.clearConnectionTimeoutHandler = function () {
+  if (this.connectionTimeoutHandler) {
+    clearTimeout(this.connectionTimeoutHandler)
+    delete this.connectionTimeoutHandler
+  }
+}
+
+/**
  * Properly setup a stream event handlers.
  *
  * @api private
@@ -969,6 +979,7 @@ Client.prototype.cleanupTimers = function () {
 Client.prototype.closeStream = function () {
   if (this.stream !== null) {
     this.stream.destroy()
+    this.clearConnectionTimeoutHandler()
     this.stream = null
   }
   if (this.connected === true || this.closed === true) {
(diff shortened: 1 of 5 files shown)


lib/nats.js
```diff
@@ -713,6 +713,11 @@ Client.prototype.setupHandlers = function () {
     return
   }
 
+  if (this.connectionTimeoutHandler) {
+    clearTimeout(this.connectionTimeoutHandler)
+    delete this.connectionTimeoutHandler
```
derekcollison: Does that make the test above fail? Meaning is that equivalent of = null?

test/tls.js
```diff
@@ -116,6 +116,26 @@ describe('TLS', () => {
     })
   })
 
+  it('should not timeout when it tls connects', done => {
```
derekcollison: connects or reconnects?

lib/nats.js
```diff
@@ -713,6 +713,11 @@ Client.prototype.setupHandlers = function () {
     return
   }
 
+  if (this.connectionTimeoutHandler) {
```
kozlovic: To me, this looks like when the timeout is originally set (when connection is created), it will be reset when processing the INFO if we switch to TLS. Which means that the timeout will be possibly increased. I would simply have the check below include that the timer be set if this.connectionTimeoutHandler is "nil" (undefined).

test/tls.js
kozlovic: Is that a test for the original issue? You set timeout of 1sec with TLS connection, but how do we check here that it would be failing (if the bug was there)? With the bug what would be the state of the connection, connected, disconnected, none of the above? Should we create a sub and from another connection (without timeout) send a message after the timeout that would have cause the disconnect, and make sure it is received?

lib/nats.js
aricart: > Does that make the test above fail? Meaning is that equivalent of = null?

clear cancels the timeout - the issue was that one of the timers was left running and when the tls upgrade happened it would create a new timer. The new timer would be fine, but the old-timer would trigger. The subtle thing is on line 723 below, which is that when the timeout function fired, it destroyed the non-tls stream, so that left it in an inconsistent state (close was not raised on the socket, so it wouldn't reactivate reconnect).

aricart: The test is effectively waiting longer than the timeout, and then chops the connection at the socket level, we are just wanting to see the reconnect happen at that point.

test/tls.js
aricart: it was on the connect - the first timer set.

lib/nats.js
aricart: The timeout is cleared at that point, and a new one created. The old timeout will call destroy on the wrong socket. This is the correct syntax for the test.

test/tls.js
aricart: just added a flush, which should accomplish the same.

lib/nats.js
kozlovic: But my point was that when setupHandlers() is called a second time, `this.stream` has been recreated (for the tls client connection) and so we the timer fires (created for the original tcp connection), we still destroy this.stream, so that should apply to the current one. Otherwise, I don't understand Node and what's the point of "this.stream" if it references a closure that was set previously.. that would be very confusing to me.

lib/nats.js
```diff
@@ -713,9 +713,13 @@ Client.prototype.setupHandlers = function () {
...
 
-  if (this.options.timeout) {
+  // tls upgrade will re-bind the handlers here, so if we have a timer we are going to ignore it
+  if (this.options.timeout && !this.connectionTimeoutHandler) {
     this.connectionTimeoutHandler = setTimeout(() => {
-      stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
+      if (this.stream) {
+        // this fires on the current stream
+        this.stream.destroy(new NatsError(CONN_TIMEOUT_MSG, CONN_TIMEOUT))
+      }
```
kozlovic: As we discussed, should we also delete timer on close stream?

lib/nats.js
derekcollison: Should we set the conncetionTimeoutHandler to nul?

CaptainYarb:
There might need to be a [test for ensuring timeouts & reconnects both work](https://gist.github.com/CaptainYarb/9240e5126958b74eec60f556fb24e52c). That said after running our own testing with this version of the nats client in this PR we were able to get consistent reconnects.

Looking forward to getting this merged. We have have thousands of nats clients and having the clients recover themselves will be a nice improvement! 😓

aricart:
> There might need to be a [test for ensuring timeouts & reconnects both work](https://gist.github.com/CaptainYarb/9240e5126958b74eec60f556fb24e52c). That said after running our own testing with this version of the nats client in this PR we were able to get consistent reconnects.
> 
> Looking forward to getting this merged. We have have thousands of nats clients and having the clients recover themselves will be a nice improvement! 😓

@CaptainYarb You tested with the ones on the current PR or the first version?  - nats@next is currently hosting the previous one, I am working out the kinks and tests on the PR listed here.

lib/nats.js
aricart: yes I refactored it to its own function

lib/nats.js
aricart: the delete does that

CaptainYarb:
> > There might need to be a [test for ensuring timeouts & reconnects both work](https://gist.github.com/CaptainYarb/9240e5126958b74eec60f556fb24e52c). That said after running our own testing with this version of the nats client in this PR we were able to get consistent reconnects.
> > Looking forward to getting this merged. We have have thousands of nats clients and having the clients recover themselves will be a nice improvement! 😓
> 
> @CaptainYarb You tested with the ones on the current PR or the first version? - nats@next is currently hosting the previous one, I am working out the kinks and tests on the PR listed here.

`nats@1.4.9-0` from `nats@next`

PR #345, opened 2020-04-17
update deps and added test

aricart:
updated dev dependencies
added test to verify that on cert error, client redials

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -886,6 +886,7 @@ Client.prototype.sendConnect = function () {
  * @api private
  */
 Client.prototype.createConnection = function () {
+  this.emit('dialing')
   this.pongs = this.pongs || []
   this.pending = this.pending || []
   this.pSize = this.pSize || 0
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -886,7 +886,6 @@ Client.prototype.sendConnect = function () {
  * @api private
  */
 Client.prototype.createConnection = function () {
-  this.emit('dialing')
   this.pongs = this.pongs || []
   this.pending = this.pending || []
   this.pSize = this.pSize || 0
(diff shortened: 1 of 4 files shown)


lib/nats.js
```diff
@@ -886,6 +886,7 @@ Client.prototype.sendConnect = function () {
  * @api private
  */
 Client.prototype.createConnection = function () {
+  this.emit('dialing')
```
derekcollison, line 889: We freeform these all the time or are they constants somewhere?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nats",
-  "version": "1.4.9-0",
+  "version": "1.4.9-2",
```
derekcollison, line 3: The -N for betas?

lib/nats.js
aricart: these are only for testing - we document all the ones that are official configuration/event properties.

package.json
aricart: yes

lib/nats.js
derekcollison: But these will be in production too and users can start to depend on them yes?

aricart: we are not responsible for users digging into the source and making dependencies on non-public features. This is simply used by a test to confirm that a dial took place when it is aborted by a tls certificate error.

derekcollison: Understood but they can, lets just make it a constant. Don't have to document it.

kozlovic: Would there be a way to use existing connection field that are updated in this process, for instance `reconnects` (I understand that this test fails during initial connect, but wonder if node is using the same logic than if it was reconnecting even on initial failed connect).

aricart: yes hooking on `reconnecting` works perfect. - will add constants for events in a different PR. this way we can group them under `Events` and `TestEvents` to further document the intention..

PR #351, opened 2020-05-13
Jitter

aricart:
Added the ability to specify jitter parameters. The jitter is an extra amount of time that is random up to specified value which is added to the standard `reconnectTimeWait`. Two values are exposed `reconnectJitter` and `reconnectJitterTLS`. In addition, there's a `reconnectDelayHandler` where a client can specify a value (in milliseconds) that will be used for delay between reconnects.


README.md
```diff
@@ -496,6 +499,23 @@ The following is the list of connection options and default values.
...
+
+The intention of the settings is to spread out the number of clients attempting to reconnect to a server over a period of time, 
+and thus preventing a ["Thundering Heard"](https://docs.nats.io/developing-with-nats/reconnect/random).
+
+The relationship between these is:
+
+- If `reconnectDelayHandler` is specified, the client will wait the value returned by this function. No other value will be taken into account.
+- If the client specified TLS options, the client will generate a number between 0 and `reconnectJitterTLS` and add it to
+`reconnectTimeWait`.
+- If the client didn't specify TLS options, the client will generate a number between 0 and `reconnectJItter` and add it to `reconnectTimeWait`.
```
kozlovic, line 516: case `reconnectJitter` instead of `reconnectJItter`.

index.d.ts
```diff
@@ -78,17 +78,26 @@ export interface ClientOpts {
 	pingInterval?: number,
 	preserveBuffers?: boolean,
 	reconnect?: boolean,
+	// jitter for non-tls connections
+	reconnectJitter?: number,
+	// jitter for tls connections
+	reconnectJitterTLS?: number,
+	// handler for custom tls connection
```
kozlovic, line 85: not sure about this comment: "// handler for custom tls connection"

README.md
aricart: good catch.

index.d.ts
aricart: removed these comments as the README file contains the doc for this.

README.md
aricart: Fixed

PR #360, opened 2020-06-15
GitHub CI

aricart:
moved from travis to github actions


derekcollison:
I would have @ripienaar take a look since he did the nats-server GHA stuff.

.github/workflows/natsjs.yml
```diff
@@ -0,0 +1,48 @@
...
+  push:
+    tags:
+      - '*'
+    branches:
+      - '*'
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
```
kozlovic, line 13: Since it is about test, I would have named this "test" instead of "build".

.github/workflows/natsjs.yml
```diff
@@ -0,0 +1,48 @@
...
+    name: ${{ matrix.config.kind }} ${{ matrix.config.os }}
```
kozlovic, line 14: Doesn't look like the name is actually used in the list [here](https://github.com/nats-io/nats.js/actions/runs/136309302). In the NATS Server, we don't have name here, and what is displayed is the name of the job "test" followed by the Go version, for instance "test 1.13", without us having to add anything. Maybe that would work here too?

.github/workflows/natsjs.yml
```diff
@@ -0,0 +1,48 @@
...
+    strategy:
+      matrix:
+        node-version: [10.x, 14.x]
+
+    steps:
+      - name: Checkout NATS.js
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Use Node.js ${{ matrix.node-version }}
```
kozlovic, line 26: see below note about `matrix.node-version`

.github/workflows/natsjs.yml
```diff
@@ -0,0 +1,48 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
```
kozlovic, line 29: In doc I see `matrix.node` instead of `matrix.node-version`. Maybe they both work, but using https://github.com/nektos/act, this is choking on that variable:
```
ERRO[0000] Unable to interpolate string '${{ matrix.node-version }}' - [ReferenceError: 'version' is not defined] 
```
whereas with `${{ matrix.node }}` it works (I am running with dry mode: act -n).

ripienaar:
The thing about these is that you make about 50 commits till it works :P

Enable actions on your fork of this repo and iterate it will run them on every push. Once it works merge it :) thats  about the best advise I can give hehe

aricart:
> The thing about these is that you make about 50 commits till it works :P
> 
> Enable actions on your fork of this repo and iterate it will run them on every push. Once it works merge it :) thats about the best advise I can give hehe

Correct - this is what I had to do for several repos now - the docs from github actions is challenging. The lack of restarting a particular build in a build matrix is close to a deal-breaker, but today, travis stopped showing the butting to restart one of the builds for some reason, so GA is the thing.

.github/workflows/natsjs.yml
aricart: Their own doc uses it. 

https://help.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions#starting-with-the-nodejs-workflow-template

kozlovic: Like I said, maybe both work, but nektos/act did not like that, and their own doc seem indeed to have both versions, the one I saw was using matrix.node [here](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-running-with-more-than-one-version-of-nodejs)

PR #361, opened 2020-06-15
added assertions to publish/subscribe/request api arguments

### LINKED ISSUE #357: Subscribe to a null topic generates a reconnect loop
If I try to subscribe to a null or undefined topic, nats client will enter in a reconnect loop.

Nats server will output the following error `processSub Parse Error: '5'`

No error is emitted by the nats client.

Using latest node client. Server version version: 2.1.7

aricart:
added assertions to publish/subscribe/request API arguments
FIX #357

### DIFF
--- a/lib/nats.js
+++ b/lib/nats.js
@@ -80,12 +80,14 @@ const PONG_RESPONSE = 'PONG' + CR_LF
 
 // Errors
 const BAD_AUTHENTICATION = 'BAD_AUTHENTICATION'
+const BAD_CALLBACK = 'BAD_CALLBACK'
 const BAD_CREDS = 'BAD_CREDENTIALS'
 const BAD_JSON = 'BAD_JSON'
 const BAD_MSG = 'BAD_MSG'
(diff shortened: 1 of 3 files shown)


lib/nats.js
```diff
@@ -1709,15 +1713,17 @@ Client.prototype.publish = function (subject, msg, reply, callback) {
...
-    // FIXME, check exists, error otherwise..
-    qgroup = opts.queue
-    max = opts.max
+Client.prototype.subscribe = function (subject, options, callback) {
+  const args = callbackShifter(3, Array.prototype.slice.call(arguments))
+  subject = args[0]
+  options = args[1] || {}
+  callback = args[2]
+
+  const qgroup = options.queue
```
kozlovic, line 1722: I really don't like node.. what is that? Meaning that above you just say: `options = args[1] || {}`, but here assume that options is not undefined and you can access options.queue (and .max).

aricart: a better version of `args[1] ? args[1] : {}` - if the `args[1]` is flasy it sets to empty object.

PR #403, opened 2021-03-03
- [update] downstream nbc changes related to constant naming

aricart:
- [breaking] constants for ErrorCode, Events, DebugEvents are now camel case as per ES5
- [change] the MsgHdrs.code property is now guaranteed to be set. Note that 0 is a successful code if the server didn't set it. - `MsgHdrs.hasError` will be true if a code value is not zero or is < 200 or ≥ 300.

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,10 +446,10 @@ try {
   console.log(m.data);
 } catch (err) {
   switch (err.code) {
-    case ErrorCode.NO_RESPONDERS:
+    case ErrorCode.NoResponders:
       console.log("no one is listening to 'hello.world'");
       break;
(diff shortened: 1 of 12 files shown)


derekcollison (commented):
Could we keep old ones with same values to not break clients?

aricart:
> Could we keep old ones with same values to not break clients?

The new API organizes those errors differently, so this change doesn't matter.

kozlovic (approved):
LGTM. @derekcollison talking with Alberto: the main branch is based on v2 that is not yet released, so this is not breaking anybody using the latest release.

PR #407, opened 2021-03-14
fix examples showing `headers` connect option

### LINKED ISSUE #405: I can not implement the new concept of Headers in my client connecting via nats.js 2.0.0-220
Hi there,
I am having some difficulties being able to implement and run your new concept of Headers, despite of the documentation and examples you have provided.

I am running a docker-compose cluster of 3 nats-server ver. 2.1.9 and I am connecting to the cluster via nats.js ver. 2.0.0-220 connector. As per your documentation: "If the client wishes to send headers, it has to enable it must add a headers field with the true value in its CONNECT message: "
In TypeScript you don't even have the option available (headers: true), however you can force it if you declare the type any instead of ConnectionOptions.

Question: Is connector nats.js ver. 2.0.0-220 ready to use Headers? Is the latest nats-server 2.1.9 published on docker-nats able to handle the protocol with headers enabled?
Thank you.

aricart:
[fix] fixed examples showing header option used to set a connection option which was removed
FIX #405

### DIFF
--- a/examples/nats-pub.js
+++ b/examples/nats-pub.js
@@ -30,10 +30,6 @@ const payload = argv._[1] || "";
 const count = (argv.c === -1 ? Number.MAX_SAFE_INTEGER : argv.c) || 1;
 const interval = argv.i || 0;
 
-if (argv.headers) {
-  opts.headers = true;
-}
-
 if (argv.debug) {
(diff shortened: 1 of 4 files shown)


kozlovic (approved):
LGTM

PR #422, opened 2021-04-26
fix honoring of `rejectUnauthorized`

### LINKED ISSUE #421: tls `rejectUnauthorized` is not honored by the client, because the client is performing a check if the socket is authorized.
v2 of the library adds a secureConnectListener to the TLS upgrade process, the `secureConnect` event is called, but the library is checking if the sockets is `authorized` which in the case of `rejectUnauthorized` won't be as the certificate signature cannot be validated against known CAs

aricart:
[fix] `rejectUnauthorized` was not getting honored due to a check to see if the connection was authorized (CA signature was verified). By passing this check if `rejectUnauthorized` is set to `false`.

FIX #421

### DIFF
--- a/src/node_transport.ts
+++ b/src/node_transport.ts
@@ -205,7 +205,11 @@ export class NodeTransport implements Transport {
 
   async startTLS(): Promise<TLSSocket> {
     let tlsError: Error;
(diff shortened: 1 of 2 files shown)


kozlovic (approved):
LGTM

PR #456, opened 2021-09-27
added explicit check on result from extractProtocolMessage

aricart:
- Added explicit test to see if the peek yielded a protocol message and not an empty string


kozlovic (approved):
LGTM

PR #458, opened 2021-10-05
[FEAT] [NBC] dns-resolve

aricart:
The node client will now try to resolve hostnames itself and attempt to connect to any of the possible alternatives before trying a different named server. Resolved ips are shuffled by default unless `noRandomize` was set in the connection options

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "nkeys.js": "^1.0.0-9"
       },
       "devDependencies": {
-        "@types/node": "^14.14.34",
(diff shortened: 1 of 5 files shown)


derekcollison (commented):
Why opt-in vs default behavior?

aricart:
> Why opt-in vs default behavior?

If we are doing semver, and make default is a change which to disable would require code - so this is a major version change. I am OK flip it.

kozlovic (approved):
LGTM

PR #471, opened 2022-01-17
Use import for code examples in the readme

Rauno56:
Readme very conveniently includes code snippets most of which have both:

1. top-level await(TLA) which is not supported outside of ESM,
2. `require` calls which is not supported **in** ESM.

So essentially the examples work nowhere in JS land(TS has TLA from v3.8).
Why not use import statements which are supported in TS and Node.js 12.20, 14.14, or 16.0 and up?


aricart (approved):
LGTM

aricart:
@Rauno56 thanks for your contribution.

PR #473, opened 2022-01-19
2.5.0

aricart:
[UPDATE] nbc to v1.5.0, build deps and GA nats-server/deno
[FIX] changed test call of `jsm.streams.update()` to use new signature
[fmt] ran latest version of the formatter


kozlovic (approved):
LGTM

PR #474, opened 2022-01-19
tls cert updates for tests on node 17

aricart:
[CI] added node 17 to CI test matrix
[UPDATE] test certs for node 17, had to disable the noiptls test until the correct cert is generated.


kozlovic (approved):
LGTM

Issue #521, opened 2022-10-10
v2.8.0 is broken on Node 10


icebob:
Based on package.json the version 2.8.0 still supports Node 10, but it contains breaking change and not working on Node 10.

![image](https://user-images.githubusercontent.com/306521/194937496-6350ddfe-86cc-426f-890f-7a567dc8c047.png)

aricart:
Actually, min requirement is Node 14 as per https://github.com/nodejs/Release

icebob:
yeah, but if you develop a library, and you supported Node 10 and drop it, you should release a new major version. 
In package.json the minimum node version is 10:

https://github.com/nats-io/nats.js/blob/e7b18d392a6da0158dc4dffce59fac94096f47b9/package.json#L57

aricart:
Since you are in node, can you add https://www.npmjs.com/package/globalthis, this will likely keep you running - Some of the APIs will not work (related to ObjectStore) etc.

aricart:
Note that references to `globalThis` have been present since 6/6/20.

PR #522, opened 2022-10-10
resetting nbc to point to main

aricart:
[BUMP] min runtime to be node 14
Also recording that min runtime supported is based on https://github.com/nats-io/nats.js#supported-node-versions


aricart:
FIX #521

PR #564, opened 2023-03-11
[FIX] added guard to see if it fixes  #560

### LINKED ISSUE #560: Errors on socket end, `Cannot read properties of undefined (reading 'end')`
We recently upgraded from version 1 to the latest version 2 of the NATS.js client, and over the last few weeks have been seeing a growing number of the following issue being reported in our Sentry:

```
Cannot read properties of undefined (reading 'end')
```
With the following stack:

Raw:
```
TypeError: Cannot read properties of undefined (reading 'end')
  File "<snip>/node_modules/nats/lib/src/node_transport.js", line 274, col 29, in <anonymous>
    this.socket.end();
  File "node:internal/streams/writable", line 500, col 5, in afterWrite
  File "node:internal/streams/writable", line 480, col 7, in onwrite
  File "node:internal/stream_base_commons", line 87, col 12, in WriteWrap.onWriteComplete [as oncomplete]
  File "node:_tls_wrap", line 581, col 11, in Socket.done
  File "node:events", line 628, col 26, in Object.onceWrapper
  File "node:events", line 513, col 28, in Socket.emit
  File "node:domain", line 489, col 12, in Socket.emit
  File "node:net", line 322, col 12, in TCP.<anonymous>
```

As this happens,  wee see the `reconnecting` event from `connection.status()` fire, and it's also sometimes accompanied with a `NatsError: 'Stale Connection'` error.

The issue itself here doesn't seem to impact anything most of the time, and the client reconnects without issue. But we have sometimes seen cases where the client hangs and requires us to completely restart it.

aricart:


codegangsta:
LGTM 👍

codegangsta (approved):

PR #565, opened 2023-03-23
fix-507

aricart:
Added pointers to the location of the nats-base-client and related documentation to the top of the file.

[ci skip]


scottf (approved):
LGTM

PR #580, opened 2023-08-01
[FEAT] added noop implementation for discard() api

aricart:
Transport interface added a `discard()` method, to allow for websocket discards on transports that throttle, this is a noop on nodejs tcp


derekcollison (approved):
LGTM

PR #586, opened 2023-08-21
Fix issue form discussion link

bruth:
Since this is derived from nats.deno, I set the discussion link to that unless you prefer otherwise.


aricart:
@bruth I think they should be separate - some of the node issues are more specific.

aricart (commented):
[skip ci]
LGTM

aricart (approved):
LGTM

PR #599, opened 2023-10-17
[CI] npm with provenance

aricart:


.github/workflows/npm.yml
```diff
@@ -0,0 +1,45 @@
...
+  release:
+    types: [created]
+
+jobs:
+  test:
+    strategy:
+      matrix:
+        node-version: [20.x]
+        deno-version: [1.36.4]
+
```
derekcollison, line 13: Bun?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nats",
-  "version": "2.17.0",
+  "version": "2.17.1-1",
```
derekcollison, line 3: Why the dash?

.github/workflows/npm.yml
aricart: not in this one.

package.json
aricart: Because that is a test version - will only make it to nats@next - that is a qualifier.

.github/workflows/npm.yml
aricart: not in this one.

derekcollison (approved):
LGTM

PR #615, opened 2024-04-08
[FEAT] enable use of `handshakeFirst` option (tlsFirst) on node client

aricart:
[BUMP] copied tls certs from nats.go


scottf (approved):
LGTM

PR #641, opened 2024-11-05
[FEAT] Bump Typescript target to Node 14 (ES2020)

Ginden:
This change:

* Improves performance (native async/await is faster than `__awaiter`)
* Improves debugging (error stacktraces are preserved through `async/await`)
* Reduces bundle size a bit (no `var __awaiter` in generated code)

This should be a safe change, as Node 14 is already required.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES6",
+    "target": "ES2020",
     "module": "CommonJS",
     "outDir": "lib/",
     "moduleResolution": "node",


aricart:
@Ginden the current in dev version bumps it to esnext 

https://github.com/nats-io/nats.js/blob/cd48c792668af0de6689b104a5efce2dbe5841a4/transport-node/tsconfig.json#L3

Ginden:
> also the min requirement for the runtime is currently Node 18

Well, it should be updated in `package.json` then - https://github.com/nats-io/nats.node/blob/main/package.json#L63

aricart:
This is fixed in https://github.com/nats-io/nats.js/blob/799117927d3c9201fbb93cf43439f94d390d1884/transport-node/package.json#L55 - part of the `@nats-io` modules - see https://github.com/nats-io/nats.js

